PR #3, opened 2023-03-17
feat: Generate OneApp XML

DafyddLlyr:


src/types.ts
```diff
@@ -72,3 +72,105 @@ export type Breadcrumbs = {
 };
 
 export type OrderedBreadcrumbs = Array<EnrichedCrumb>;
+
+export interface Passport {
+  data?: Record<string, any>;
```
builditben, line 77: As discussed, it would be nice to eventually refactor this to a tighter definition (i.e. use the `Value` type introduced in #4 ) but not necessary for this change.

src/types.ts
```diff
@@ -72,3 +72,105 @@ export type Breadcrumbs = {
...
+export interface GovUKPayment {
+  amount: number;
+  reference: string;
+  state: {
+    status: PaymentStatus;
+    finished: boolean;
+  };
+  payment_id: string;
+  created_date: string;
+  _links: {
```
builditben, line 103: nit: is `_links` necessary to include in this interface? I think this is only used in the editor during a live session but not after that.

src/types.ts
```diff
@@ -72,3 +72,105 @@ export type Breadcrumbs = {
...
+  pao?: string;
+  street?: string;
+  town?: string;
+  postcode?: string;
+  single_line_address?: string;
+  planx_description?: string; // joined via table blpu_codes
+  planx_value?: string; // joined via table blpu_codes
+}
+
+export const GOV_PAY_PASSPORT_KEY = "application.fee.reference.govPay" as const;
```
builditben, line 163: nit: it would be nice to eventually collect a dictionary of meaningful passport keys in its own file but that's a future improvement.

src/passport.ts
builditben, line 1: Really nice to see the first proper domain class being introduced here :heavy_plus_sign: :tada:

builditben (approved):
Nice PR. It's really good to see the direction we're going in here towards a more declarative API. Generating the OneAppXML from just a sessionId is ideal :ok_hand:

builditben:
@DafyddLlyr sorry, I just saw that this is still a draft. I think you're just adding more tests but if you also want me to take another look, just let me know.

src/types.ts
DafyddLlyr: Rebased and pulled this type in 👌

DafyddLlyr: This turns out to be a bit more complex than originally thought - multiple functions inside the OneApp class rely on being passed a certain type and not a `Value`.

In order to do this I'll need to get a `Passport().get()` function in place which I'll revisit as a discrete PR shortly in order to move forward on this sprint's tickets (send templates to email).

src/types.ts
DafyddLlyr: This is a fair question - we'll need to think about where and how we want to handle types across the repos. I've kept it in full here for now for simplicity but agree that this part of the model is an implementation concern.

PR #5, opened 2023-03-22
feat: Add StaticSessionState with sectionOverview method

builditben:
This PR adds `StaticSessionState` which loads a flow and provides an interface for querying the state of a given set of `OrderedBreadcrumbs`. This includes a [`sectionOverview` method](https://github.com/theopensystemslab/planx-core/compare/ben/section-statuses?expand=1#diff-4e338f99f93be240455fd092cf30f386989431ed3357c853fb53a46f64a5d212R90-R171) which constructs a `SectionOverview` object representing the current state of the session.

`OrderedBreadcrumbs` are a replacement for `EnrichedBreadcrumbs`  since the collection of `NormalizedCrumb` objects have enough information for logical operations while the additional node details can be trivially looked-up when needed. `sortBreadcrumbs` is added back in with this change.

This PR lays the groundwork for a standalone state machine for breadcrumbs which should provide a cleaner split between data logic and presentation in the editor and API.


src/types.ts
```diff
@@ -66,7 +66,6 @@ export interface IndexedNode extends Node {
 export type OrderedFlow = Array<IndexedNode>;
 
 export interface NormalizedNode extends IndexedNode {
-  component: string;
```
builditben, line 69: I have removed this since a lookup of the enum directly is better in code (i.e. `ComponentType.Question` instead of "Question")

For the stored data structure, serializing this as a string would be the nicest way to handle this in the future since the value would then also be human readable. We may also want to transition `ComponentType` to use an object "as const" instead of a numeric enum.

src/types.ts
```diff
@@ -85,22 +84,32 @@ export type Breadcrumbs = {
...
   sectionId?: string;
-  details: {
-    component: string;
-    nodeData?: Record<string, Value>;
-    answerData?: {
-      [id: NodeId]: Record<string, Value>;
-    };
-  };
+  rootNodeId: string;
 }
```
builditben, line 93: Enough information has been added to allow an array of `NormalizedCrumb`s to be easily traversed for queries but the data from each node can be looked-up separately (that's why this has changed from `EnrichedCrumb`).

src/session-state.test.ts
DafyddLlyr, line 1: Tests here are super clear and meaningful - solid documentation of the intended functionality.

src/session-state.ts
```diff
@@ -0,0 +1,187 @@
...
+  FlowGraph,
+  NormalizedFlow,
+  NormalizedNode,
+  SectionOverview,
+} from "./types";
+
+// TODO
+// SessionState loads a flow and provides an interface for
+// creating, updating and exporting a set of breadcrumbs
+// export class SessionState {}
```
DafyddLlyr, line 17: 👍

src/session-state.ts
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  remainingNodes(breadcrumbs: OrderedBreadcrumbs): Array<NormalizedNode> {
+    if (breadcrumbs.length == 0) {
+      return this.normalizedFlow;
+    }
+    const currentIndex = this.currentIndex(breadcrumbs);
+    const currentNode = this.normalizedFlow[currentIndex]!;
+    return this.normalizedFlow.filter((node, index) => {
+      const afterCurrent = index > currentIndex;
```
DafyddLlyr, line 82: I may well have missed this somewhere but how are we thinking of handling checklists which make a journey "revisit" another (earlier) forking point?

src/session-state.ts
```diff
@@ -0,0 +1,187 @@
...
+      const nextChildNode = afterCurrent && node.parentId == currentNode.id;
+      const nextRootNode =
+        afterCurrent && node.rootNodeId != currentNode.rootNodeId;
+      return nextChildNode || nextRootNode;
+    });
+  }
+
+  sectionOverview({
```
DafyddLlyr, line 90: This is great 👌

DafyddLlyr (approved):
Quite a lot to take in here in terms of functionality, but the style is great - super clear and succinct. Looking forward to being able to use a lot of this in practice.

Only one real question which I'm not sure we've discussed, which is the concept of checklists or traversing multiple branched where we essentially need to "go back" up the chain of nodes.

src/session-state.ts
builditben: @DafyddLlyr thanks - I'd love to pick this up in more detail on the dev call (if possible). I do have a `previousNodes` method on another branch but I suspect you're talking about a non-obvious behavior in going forwards. Happy to add some more test cases around this as we go to cover these kinds of scenarios.

src/types.ts
jessicamcinchak: sorry to be catching up here after this merged, but a couple questions coming up for me as I'm reading thorugh these changes:
- what exactly is `rootNodeId` here & on `NormalizedFlow`?
- how is it different than `id` ?

PR #6, opened 2023-03-22
feat: GetTemplates for flow and session

DafyddLlyr:
When handling templates, we currently derive everything from a `sessionId`. A new function has been added to allow us to fetch template names from a `sessionId`.

Relies on db changes made in https://github.com/theopensystemslab/planx-new/pull/1553


src/index.ts
```diff
@@ -75,8 +75,17 @@ export class CoreDomainClient {
     return publishFlow(this.client, args);
   }
 
+  // TODO: Remove this once planx-new updated
```
jessicamcinchak, line 78: :+1:

jessicamcinchak (approved):

PR #7, opened 2023-03-27
feat: Add `createPaymentRequest` and `lockSession` methods

builditben:


src/payment-request.ts
```diff
@@ -0,0 +1,77 @@
...
+): Promise<PaymentRequest> {
+  const isSessionLocked = await checkSessionLock(client, sessionId);
+  if (!isSessionLocked) {
+    throw new Error(
+      "session mush be locked before a payment request can be created"
+    );
+  }
+  const session = await getSessionById(client, sessionId);
+  const sessionPreviewData: PaymentRequest["sessionPreviewData"] =
+    buildSessionPreviewData(session, sessionPreviewKeys); // throws if sessionPreviewData cannot be built
```
builditben, line 29: Building the session preview data from keys ensures that the data exists on the session. The client can specify the keys but the API doesn't need to trust client data directly.

src/types.ts
```diff
@@ -209,7 +213,19 @@ export interface SessionData {
 }
 
 export interface Session {
-  data: SessionData,
+  data: SessionData;
   id: string;
   flowId: string;
 }
+
+export interface DetailedSession extends Session {
```
builditben, line 221: It would be nice to eventually land on a naming convention for data transfer objects (i.e. objects with database fields) vs. domain objects (i.e. the data model) vs utility objects (i.e. unusually shaped objects from custom queries). "Detailed" as a prefix for data transfer objects feels fine for now IMO but suggestions welcome.

DafyddLlyr: I also personally prefer meaningful types with names, sometimes using the utility types `Pick<T>` and `Omit<T>` rather than just adding optional properties to types which can happen a lot for convenience. Going for a new named type here is the right move I think.

DafyddLlyr (approved):
Great! Super clean and simple (in the best of ways!)

PR #8, opened 2023-03-27
chore: Move types to be importable from "@opensystemslab/planx-core/types"

builditben:
This is a proposal to move the definition of shared type definitions into their own folder where they can be exposed. 

This enables usage like `import type { ... } from "@opensystemslab/planx-core/types"` and prevents having to reexport from the index which isn't a nice pattern.

Shared types have also been split into sensible groupings of individual files to make maintenance easier.


src/types/index.ts
DafyddLlyr, line 1: There's a few random types in the file that probably belong with the OneApp generation, and their corresponding `planx-new` types can be swapped out.

I'm going to revisit the Passport class shortly so I'm happy to pick this up 👍

DafyddLlyr (approved):
Much better - thank you for picking this up 🙌

PR #13, opened 2023-04-05
feat: add payment amount to payment requests

builditben:
Payment requests can only be created for flows with a pay component. This adds a lookup for the "amount" passport key from the pay component and add it to sessionPreviewKeys if it exists and fails the payment request creation if not.

### DIFF
--- a/src/flow.ts
+++ b/src/flow.ts
@@ -8,7 +8,7 @@ export async function getLatestFlowGraph(
 ): Promise<FlowGraph | undefined> {
   const { published_flows: response } = await client.request(
     gql`
-      query GetPublishedFlowData($flowId: uuid!) {
+      query GetLatestPublishedFlowData($flowId: uuid!) {
         published_flows(
           where: { flow_id: { _eq: $flowId } }
           order_by: { created_at: desc }
@@ -20,7 +20,7 @@ export async function getLatestFlowGraph(
     `,
     { flowId }
   );
-  return response.published_flows.length ? response.published_flows[0] : null;
+  return response?.length ? response[0] : null;
 }
 
 export async function createFlow(
--- a/src/flow.ts
+++ b/src/flow.ts
@@ -1,5 +1,27 @@
 import { gql } from "graphql-request";
 import type { GraphQLClient } from "graphql-request";
+import type { FlowGraph } from "./types";
+
+export async function getLatestFlowGraph(
+  client: GraphQLClient,
+  flowId: string
+): Promise<FlowGraph | undefined> {
+  const { published_flows: response } = await client.request(
+    gql`
+      query GetPublishedFlowData($flowId: uuid!) {
+        published_flows(
+          where: { flow_id: { _eq: $flowId } }
+          order_by: { created_at: desc }
+          limit: 1
+        ) {
+          data
+        }
+      }
+    `,
+    { flowId }
+  );
+  return response.published_flows.length ? response.published_flows[0] : null;
+}
 
 export async function createFlow(
   client: GraphQLClient,
--- a/src/flow.ts
+++ b/src/flow.ts
@@ -20,7 +20,7 @@ export async function getLatestFlowGraph(
     `,
     { flowId }
   );
-  return response?.length ? response[0] : null;
+  return response?.length ? response[0].data : null;
 }
 
 export async function createFlow(
--- a/src/flow.ts
+++ b/src/flow.ts
@@ -5,7 +5,7 @@ import type { FlowGraph } from "./types";
 export async function getLatestFlowGraph(
   client: GraphQLClient,
   flowId: string
-): Promise<FlowGraph | undefined> {
+): Promise<FlowGraph | null> {
   const { published_flows: response } = await client.request(
     gql`
       query GetLatestPublishedFlowData($flowId: uuid!) {
--- a/src/payment-request.test.ts
+++ b/src/payment-request.test.ts
@@ -1,7 +1,7 @@
-import { buildSessionPreviewData } from "./payment-request";
+import { validateSessionData } from "./payment-request";
 import type { Session, KeyPath } from "./types";
 
-describe("buildSessionPreviewData", () => {
+describe("validateSessionData", () => {
   test("passport data must be available", () => {
     const emptySession: Session = {
       id: "abc",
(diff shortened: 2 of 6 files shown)


src/flow.ts
```diff
@@ -1,5 +1,27 @@
...
+          order_by: { created_at: desc }
+          limit: 1
+        ) {
+          data
+        }
+      }
+    `,
+    { flowId }
+  );
+  return response?.length ? response[0].data : null;
```
jessicamcinchak, line 23: nit: can this `return response?.[0]?.data`? otherwise i think falling back to `null` fails to  align with the the function type says it'll return `Promise<... | undefined>`?

src/payment-request.ts
```diff
@@ -1,8 +1,22 @@
...
+  Session,
+  KeyPath,
+  Value,
+  Node,
+  FlowGraph,
+} from "./types";
+
+type ValidatedSessionData = {
+  sessionPreviewData: PaymentRequest["sessionPreviewData"];
+  amount: number;
```
jessicamcinchak, line 18: nit: i'd prefer to standardise around the language of `fee` as that feels more specific to the process of paying/requesting payments & better aligns to existing LDC schemas (eg `application.fee.payable`), whereas `amount` reads as broader to me and I think could be more easily confused with other things in content like number of homes on a site, amount of windows being replaced, etc!

src/payment-request.test.ts
```diff
@@ -28,15 +124,23 @@ describe("buildSessionPreviewData", () => {
...
-    expect(sessionPreviewData).toEqual({ a: 1, b: 2, c: 3 });
+    const validatedSessionData = validateSessionData({
+      session: session,
+      sessionPreviewKeys: previewKeys,
+      amountKey: "amount",
+    });
+    expect(validatedSessionData).toEqual({
+      amount: 5,
+      sessionPreviewData: { a: 1, b: 2, c: 3 },
+    });
```
jessicamcinchak, line 143: thanks for very clear tests :+1:

src/payment-request.ts
```diff
@@ -54,27 +79,69 @@ export async function createPaymentRequest(
...
+    if (node.type === ComponentType.Pay) {
+      payNode = node;
+      return;
+    }
+  });
+  if (!payNode) {
+    throw new Error("flow must contain a pay node");
+  }
+  return (payNode.data?.fn as string) || "amount";
+}
```
jessicamcinchak, line 115: is it worth adding a comment here that this is explicitly assuming one Pay node per flow and the passport key is reflecting the _first_ Pay node it finds from bottom up? (which is a totally fair assumption for now based on prod content, but nice to be explicit about in my opinion!)

src/payment-request.ts
```diff
@@ -54,27 +79,69 @@ export async function createPaymentRequest(
...
+    throw new Error("flow not found");
+  }
+  // search for pay component from the bottom of the root node upwards
+  let payNode: Node | undefined;
+  flowGraph._root.edges.reverse().forEach((rootNodeId) => {
+    const node = flowGraph[rootNodeId];
+    if (node.type === ComponentType.Pay) {
+      payNode = node;
+      return;
+    }
```
jessicamcinchak, line 109: on first read, it looks like this is only checking node types that are directly on the "root" and would fail to find a Pay component on a branch, is that right? 

LDC content looks like this & will need to account for branches! But the bottom-up search direction is clever here 
Screenshot from 2023-04-06 10-18-51

jessicamcinchak (commented):
this is looking really solid! left a few questions and naming nitpicks, happy to talk through / prioritise re-reviewing later today to move this forward soon

src/flow.ts
builditben: Weird, I would have expected Typescript to have spotted this :shrug: 
Thanks, I've fixed the type to return null as intended.

src/payment-request.ts
builditben: As mentioned above, this now uses breadcrumbs to find the relevant pay component.

DafyddLlyr (approved):
I'll have a look at making the required changes on the agent/applicant journey now to line up with this 👍

src/payment-request.ts
jessicamcinchak: This sounds like a good plan :heavy_check_mark:  

And for clarity: Sections are the only node type that we currently validate in a graph, all other node types including Pay can exist in any position in a graph and have zero or many instances. So searching breadcrumb rather than flow is definitely a good shout and the more efficient way forward based on our graph assumptions :+1:

DafyddLlyr: Update from dev call - we'll add a breadcrumb at the API level to achieve this.

src/payment-request.ts
```diff
@@ -27,54 +37,141 @@ export async function createPaymentRequest(
...
+  // the pay amount in the passport
+  for (const node of sortFlow(flowGraph)) {
+    const isNextNodeAfterLastCrumb = node.parentId === lastCrumb.id;
+    if (isNextNodeAfterLastCrumb && node.type === ComponentType.Pay) {
+      const amountKey = getPaymentAmountKeyFromNode(
+        flowGraph[node.id] as PayNode
+      );
+      return passport[amountKey];
+    }
+  }
```
builditben, line 168: This function looks up the payment amount key from either the last breadcrumb or the next breadcrumb that is also a Pay component. I could also update the breadcrumbs here but decided against it in this PR.

src/types.ts
```diff
@@ -228,8 +228,10 @@ export interface DetailedSession extends Session {
 
 export interface PaymentRequest {
   id: string;
+  applicantName: string;
   sessionId: string;
   payeeName: string;
   payeeEmail: string;
+  paymentAmount: number;
```
builditben, line 235: Applicant name and payment amount have been added.

builditben (commented):
I updated this PR after yesterday's discussion. There are a few minor changes and a larger change to how the payment amount is found.

src/payment-request.ts
builditben: I found a slightly different strategy for this PR (see below). If it's still helpful to modify breadcrumbs, that's doable too.

builditben:
[[Editor] Payee standalone journey](https://trello.com/c/uDD2GMtJ/2350-editor-payee-standalone-journey)

builditben:
[[Editor] Payee standalone journey](https://trello.com/c/uDD2GMtJ/2350-editor-payee-standalone-journey)

DafyddLlyr (approved):

PR #15, opened 2023-04-13
refactor: tidy type declarations

builditben:
This PR moves the definition of shared type definitions into their own folder where they can be exposed.

This enables usage like import type { ... } from "@opensystemslab/planx-core/types" and prevents having to re-export from the index which isn't a nice pattern.

Shared types have also been split into sensible groupings of individual files to make maintenance easier.

Implemented in corresponding planx PR: https://github.com/theopensystemslab/planx-new/pull/1666


.github/workflows/test.yml
```diff
@@ -21,31 +21,12 @@ jobs:
...
-        uses: actions/cache@v3
-        with:
-          path: ${{ steps.pnpm-cache.outputs.STORE_PATH }}
-          key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
-          restore-keys: |
-            ${{ runner.os }}-pnpm-store-
-
-      - name: Install dependencies
-        id: install-packages
-        run: pnpm install
```
builditben, line 48: I'm not sure caching is particularly valuable in this repo - the latency between steps negates the benefits at the moment.

src/document-templates.ts
```diff
@@ -5,15 +5,18 @@ export async function getDocumentTemplateNamesForFlow(
...
-  const { flow_document_templates: response } = await client.request(gql`
-    query GetDocumentTemplateNamesForFlow($flowId: uuid!) {
-      flow_document_templates(where: {flow: {_eq: $flowId}}) {
-        document_template
-      }
-    }`,
-    { flowId }
-  );
-  return response.map(
+  const response: { flow_document_templates: [{ document_template: string }] } =
```
builditben, line 8: The type declarations are updated here to properly unpack the returned `unkown` type returned by the updated GraphQLClient.

src/index.ts
```diff
@@ -11,34 +11,12 @@ import {
...
   getDocumentTemplateNamesForSession,
 } from "./document-templates";
-import { Session, PaymentRequest, KeyPath } from "./types";
+import type { Session, PaymentRequest, KeyPath } from "./types";
 
 const defaultURL = process.env.HASURA_GRAPHQL_URL;
 
 export * from "./passport";
 export * from "./logic";
-export { StaticSessionState } from "./session-state";
```
builditben, line 20: `StaticSessionState` is still experimental so I'm no longer exporting it here

src/types/index.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export * from "./data";
+export * from "./component";
+export * from "./gov-uk-payment";
+export * from "./flow";
+export * from "./sections";
+export * from "./breadcrumbs";
+export * from "./component";
+export * from "./payment-request";
+
```
builditben, line 12: There are still some types that could be organized below but I've left them for future updates once sensible groupings emerge (feel free to update these if you see a grouping that makes sense).

package.json
```diff
@@ -14,48 +14,57 @@
...
     "test:coverage": "jest; open ./coverage/lcov-report/index.html",
     "check": "tsc --project ./tsconfig.check.json && pnpm lint",
     "prepare": "pnpm i rimraf && pnpm build"
   },
   "dependencies": {
-    "fast-xml-parser": "^4.1.3",
+    "fast-xml-parser": "^4.2.2",
     "graphql": "^16.6.0",
-    "graphql-request": "^5.1.0",
+    "graphql-request": "^5.2.0",
```
builditben, line 40: This update includes a type change which meant a few type fixes.

src/types/index.ts
```diff
@@ -0,0 +1,75 @@
...
+  pao?: string;
+  street?: string;
+  town?: string;
+  postcode?: string;
+  single_line_address?: string;
+  planx_description?: string; // joined via table blpu_codes
+  planx_value?: string; // joined via table blpu_codes
+}
+
+export const GOV_PAY_PASSPORT_KEY = "application.fee.reference.govPay" as const;
```
jessicamcinchak, line 58: nit: this const could be grouped within `gov-uk-payment` type file?

jessicamcinchak (approved):

src/types/component.ts
```diff
@@ -0,0 +1,33 @@
+// This needs to stay in sync with the TYPES enum
+// in planx-new/editor.planx.uk/src/@planx/components/types.ts
```
DafyddLlyr, line 2: I guess this means a next step would be to remove this from `planx-new`..? There's probably still a few duplicated types we can look to tidy up I'd guess.

DafyddLlyr (approved):
Thanks for tackling this one - it's going to really help in taking things forward.

PR #18, opened 2023-04-18
fix: Allow optional email and telephone for applicants

DafyddLlyr:
A recent Uniform submission has failed as applicant contact details were missing (email and telephone).

Our LDC service allows this, but the OneApp schema (and our Zod schema which represents it) does not.

This change allows applicant contact details to be optional, as long as they are supplied for an agent. I'm not totally sure if Uniform is validating this - I suspect it's not an issue but just something that might still cause a failure following this change.

### DIFF
--- a/src/export/oneApp/OneApp.test.ts
+++ b/src/export/oneApp/OneApp.test.ts
@@ -786,7 +786,7 @@ test("Parsing error", () => {
     files: [],
   }
   expect(() => new OneAppPayload(invalidConfig).buildXML()).toThrowError(/Invalid OneApp Payload/);
-})
+});
 
 test("Unhandled error", () => {
   const config = {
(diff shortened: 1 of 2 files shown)


src/export/oneApp/OneApp.test.ts
```diff
@@ -797,4 +797,69 @@ test("Unhandled error", () => {
...
+        "applicant.agent.phone.primary": "(123) 456789",
+      },
+    };
+    expect(() => new OneAppPayload({
+      sessionId,
+      passport,
+      files,
+    }).buildXML()).not.toThrowError();
+  });
+});
```
jessicamcinchak, line 865: thanks for clear tests :+1:

jessicamcinchak (approved):

DafyddLlyr:
Holding off for content confirmation from Alastair and/or George before I merge this one, juuust in case my assumptions are incorrect.

PR #20, opened 2023-04-21
feat: Human readable project types

DafyddLlyr:
Port of logic used in PlanX API to a shared library for use in both Editor and API. Please see  https://github.com/theopensystemslab/planx-new/pull/1658 for implementation and notes on tests.

Thinking about this over the weekend - this is currently just a copy/paste of code from `planx-new` and doesn't really fit the API we're going for here atm.

I'll update this so it returns a list for a session id, with formatting as a single string being an additional function


tsconfig.check.json
```diff
@@ -3,14 +3,14 @@
     "allowJs": true,
     "esModuleInterop": true,
     "forceConsistentCasingInFileNames": true,
-    "lib": ["es2021"],
+    "lib": ["ES2021"],
```
DafyddLlyr, line 6: When this was in lower case I was hitting an error importing `Intl.ListFormat` (which was added in ES2021).

builditben: It's less than ideal that this upper-case version is correct for [`lib`](https://www.typescriptlang.org/tsconfig#lib) but [`target`](https://www.typescriptlang.org/tsconfig#target) refers to the lowercase `es2021` ! :face_exhaling:

DafyddLlyr:
Appreciate the feedback here! I've gone for a simpler approach here similar to what I initially had - makes a lot more sense I think.

PR #21, opened 2023-04-24
feat: add BOPS payload as an export type

jessicamcinchak:
goal: for a given `sessionId`, generate and return the JSON payload body expected by BOPS - connected up over here https://github.com/theopensystemslab/planx-new/pull/1671

first attempt at porting everything over from planx-new frontend, it is/was admittedly much trickier than expected ! 

key changes: 
- `generateBOPSPayload` is introduced as a method on the `CoreDomainClient` class
- in `export/bops/index.ts`, `formatProposalDetails` (fka `makePayload`) and `getBOPSParams` are still the main functions that handle the data transformation. I have intentionally not made them directly available yet on the `CoreDomainClient`, although I suspect we'll want them, or ideally some more generalised version of them, to be available sooner than later since they are currently called by `makeCsvData` in the Planx editor and used elsewhere


src/export/bops/index.ts
```diff
@@ -0,0 +1,534 @@
...
+    case ComponentType.Checklist:
+    case ComponentType.DateInput:
+    case ComponentType.NumberInput:
+    case ComponentType.Question:
+    case ComponentType.TextInput:
+    case ComponentType.ContactInput:
+      return true;
+
+    default:
+      throw new Error(`Unhandled type: ${type}`);
```
builditben, line 77: I saw a clever trick for getting typescript to show an error when a case statement isn't exhaustive (e.g. when we add a new ComponentType):
image
https://medium.com/technogise/type-safe-and-exhaustive-switch-statements-aka-pattern-matching-in-typescript-e3febd433a7a

jessicamcinchak: thanks for pointing this out! we do something similar in planx-new in a couple places using an `exhaustiveCheck` function, which was actually improperly implemented in the BOPS file over there so I didn't copy it over and hadn't updated it yet - but added back properly now :+1:

src/export/bops/bops.test.ts
```diff
@@ -0,0 +1,187 @@
...
+  });
+
+  it("omits works from the payload when neither date is present in the passport", () => {
+    const generatedPayload = getBOPSParams({
+      ...testParams,
+      passport: {},
+    });
+    expect(generatedPayload).not.toHaveProperty("works");
+  });
+});
```
jessicamcinchak, line 187: I still need to port over all the various payload tests from the frontend here - especially around file-tagging and flags. 

Does one long test file feel like the right approach here (similar to how `export/oneApp` was ported over) or any objections to making a `exports/bops/tests/` subdirectory with files per payload property more like the current editor structure? I'm leaning towards the latter I think.

src/export/bops/index.ts
```diff
@@ -0,0 +1,536 @@
...
+// const addSectionName = (
+//   id: string,
+//   metadata: QuestionMetaData
+// ): QuestionMetaData => {
+//   const { hasSections, getSectionForNode } = useStore.getState();
+//   if (hasSections) {
+//     metadata.section_name = getSectionForNode(id).data.title;
+//   }
+//   return metadata;
+// };
```
jessicamcinchak, line 121: This is a known TODO & will require bringing over a few methods from the `NavigationStore`. Since sections are not actually in use yet on LDCs and therefore not actually being sent to BOPS, I've pushed this off. I can tack it onto these changes, or pick up separately in a small followup PR.

src/export/bops/model.ts
```diff
@@ -0,0 +1,146 @@
...
+  [key: string]: LooseNode;
+};
+
+interface LooseCrumb extends Crumb {
+  data?: Record<string, any>;
+}
+
+export type LooseBreadcrumbs = {
+  [key: string]: LooseCrumb;
+}
```
jessicamcinchak, line 21: I've defined these `Loose...` types to get around some issues using the existing `Value` type - they let us get to payload-parity much quicker here, and my intention is that we'll remove them over time as we revisit/refactor/better generalise key methods like `formatProposalDetails` to use the new stricter types.

src/result.ts
```diff
@@ -0,0 +1,89 @@
...
+        responses: responses.every((r: any) => r.hidden)
+          ? responses.map((r: any) => ({ ...r, hidden: false }))
+          : responses,
+      };
+
+      return acc;
+    },
+    {} as ReturnType<ResultData>
+  );
+};
```
jessicamcinchak, line 89: this method is copied directly from the `PreviewStore`, and similar case with `src/flags`, so that we can correctly set the `result` key on the BOPS payload.

it feels especially less than ideal as this result method & flags are ripe for change with the known logic issues/tickets around filters in the queue.

should we export this from planx-core and actually replace it in the `PreviewStore`? Should we temporarily maintain it in two places until that logic work and consolidate then?

jessicamcinchak:
Noted a few known to-dos in comments, but would very much appreciate a review / sense check of this at this stage!

src/export/bops/model.ts
```diff
@@ -0,0 +1,146 @@
...
+  | "Utility Bill"
+  | "Building Control Certificate"
+  | "Construction Invoice"
+  | "Council Tax Document"
+  | "Tenancy Agreement"
+  | "Tenancy Invoice"
+  | "Bank Statement"
+  | "Statutory Declaration"
+  | "Other"
+  | "Sitemap";
```
builditben, line 137: `Sitemap` doesn't appear in [BOP's tags](https://github.com/unboxed/bops/blob/master/app/models/document.rb) - is that new?

src/export/bops/tests/applicationType.test.ts
```diff
@@ -0,0 +1,36 @@
...
+    flow: { _root: { edges: [] } },
+    passport: {},
+    sessionId: "123"
+  };
+
+  it("defaults to `lawfulness_certificate` if we can't fetch current route", () => {
+    const generatedPayload = getBOPSParams({
+      ...testParams,
+      // @ts-ignore: Type 'undefined' is not assignable to type 'string'
+      flowName: undefined,
```
builditben, line 16: The typechecker makes this test a bit redundant also we're not relying on fetching the flow name from the route anymore. Maybe we don't need this test?

src/flags.ts
```diff
@@ -0,0 +1,113 @@
...
+
+const parsedFlags: Record<FlagSet, { [id: string]: Flag }> = Object.entries(
+  categoriesAndFlags
+).reduce((acc: Record<string, any>, [category, flags]) => {
+  acc[category] = Object.entries(flags).reduce(
+    (acc: Record<string, Flag>, [id, [text, bgColor, officerDescription]]) => {
+      // loop through all the flags and add a text color which is
+      // white if it's a dark background or black if it's light
+      const color = "#000";
+        // mostReadable(bgColor, ["#000", "#FFF"])?.toHexString() || "#000";
```
builditben, line 90: I guess `color` isn't needed for BOPS, is that right? Can we remove this commented out code and comments?

src/result.ts
```diff
@@ -0,0 +1,89 @@
...
+import { DEFAULT_FLAG_CATEGORY, flatFlags } from "./flags";
+import { Breadcrumbs, ComponentType, Flag, ResultData } from "./types";
+
+const SUPPORTED_DECISION_TYPES = [ComponentType.Checklist, ComponentType.Answer];
+
+export const getResultData = (
+  breadcrumbs: Breadcrumbs,
+  flow: LooseFlowGraph,
+  flagSet: Parameters<ResultData>[0] = DEFAULT_FLAG_CATEGORY,
+  overrides?: Parameters<ResultData>[1]
```
builditben, line 11: nit: These two type declarations are pointing at the parameters of a function. Could they point at the types directly to make this more readable? Maybe more specific `ResultData` types need to be exported.

src/result.ts
builditben: A consolidation at some point in the near future sounds like a good plan :+1:

src/types/result.ts
```diff
@@ -0,0 +1,20 @@
+
+export type ResultData = (
+  flagSet?: string,
```
builditben, line 3: nit: should this be:
```suggestion
  flagSet?: FlagSet,
```

src/types/result.ts
```diff
@@ -0,0 +1,20 @@
...
+  overrides?: { [flagId: string]: { heading?: string; description?: string } }
```
builditben, line 4: nit: Could we extract this interface out and export it so that `getResultData` can refer to the same type directly?

builditben (approved):
Thanks for this huge effort. 

I've left a few comments, suggestions and optional updates. I think a round of refactoring can follow this initial change so nothing here looks like it would hold-up merging.

A few small clean-ups would be nice: I think the lint error messages for `x is defined but never used` can be addressed. Also, it would be well worth running `npx prettier -w` to reformat files in our standard style (a `lint:fix` command is coming soon).

src/types/result.ts
```diff
@@ -0,0 +1,20 @@
...
+    displayText: { heading: string; description: string };
+  };
+}
+
+export interface Flag {
+  category: string;
+  value?: string;
+  text: string;
+  bgColor: string;
+  color: string;
```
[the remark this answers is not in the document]
jessicamcinchak: They don't, this was a result of porting over from editor. I've removed all references in planx-core to `bgColor` & `color` for simplicity - we'll just have to remember to re-account for these whenever we revisit to consolidate Result & Flag methods between planx-core & planx-new

src/export/bops/index.ts
```diff
@@ -0,0 +1,536 @@
...
+    site.x = address.x;
+    site.y = address.y;
+
+    site.source = address.source; // reflects "os" or "proposed"
+
+    data.site = site;
+  }
+
+  // 1b. property boundary
+  const geojson = findGeoJSON(flow, breadcrumbs);
```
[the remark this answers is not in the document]
jessicamcinchak: You're right! This was originally written to account for DrawBoundary having a dynamic passport variable, but it doesn't in practice! Updated and removed `findGeoJSON`

src/export/bops/tests/files.test.ts
```diff
@@ -0,0 +1,235 @@
...
+    "Unmatchable passport key": {
+      key: "foo",
+      tags: [],
+    },
+    "Location plan": {
+      key:
+        PASSPORT_UPLOAD_KEY === "proposal.drawing.locationPlan"
+          ? PASSPORT_UPLOAD_KEY
+          : "key changed unexpectedly!",
+      tags: ["Proposed", /*"Drawing",*/ "Site", "Plan"],
```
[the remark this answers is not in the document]
jessicamcinchak: These are old iterations of accepted BOPS tags; we still hold old submissions that used these tags. Flags were changing frequently for a period of time which I think explains why we were commenting rather than deleting, will flag as a future cleanup task :bookmark:

src/export/bops/index.ts
```diff
@@ -0,0 +1,536 @@
...
+      return "proposed";
+  }
+};
+
+const extractFileDescriptionForPassportKey = (
+  passport: Passport["data"],
+  passportKey: string
+): string | undefined => {
+  try {
+    // XXX: check for .description or .reason as there might be either atm
```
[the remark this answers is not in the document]
jessicamcinchak: I was never really familiar with this syntax prior to working with John & Gunar but have picked it up since: I understand it as a sort of synonym to `TODO` comments, but for known caveats/tricky logic notes that aren't actually outstanding tasks. It's equally easy to globally search for and sets apart from "regular" comments.

If you don't want it polluting planx-core, we don't have to port it over from planx-new though :upside_down_face:

src/export/bops/index.ts
```diff
@@ -0,0 +1,536 @@
...
+//   metadata: QuestionMetaData
+// ): QuestionMetaData => {
+//   const { hasSections, getSectionForNode } = useStore.getState();
+//   if (hasSections) {
+//     metadata.section_name = getSectionForNode(id).data.title;
+//   }
+//   return metadata;
+// };
+
+export const formatProposalDetails = (
```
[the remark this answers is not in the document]
jessicamcinchak: Yes same. I have a long wishlist for the next version of this function :crystal_ball:, thanks for being down to treat initial porting over and refactors as separate tasks though to keep changes a bit more contained/easier to understand.

src/export/bops/model.ts
jessicamcinchak: Good catch, I think this is an outdated tag - now removed!

src/flags.ts
jessicamcinchak: Yep, I had originally kept this file the same as `planx-new` and just commented out this line to avoid adding an unnecessary package dependency here - but realise how that might be confusing. Fully removed references to `color` and `bgColor` now as those aren't required in export data formats.

src/types/result.ts
jessicamcinchak: Yep, done!

src/export/bops/index.ts
```diff
@@ -0,0 +1,536 @@
...
+          });
+        } catch (err) {
+          throw new Error(`Error formatting files: ${err}`);
+        }
+      });
+    });
+
+  // 3. constraints
+  const constraints = (
+    passport.data?.["property.constraints.planning"] || []
```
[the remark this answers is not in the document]
jessicamcinchak: This suggestion makes a ton of sense, but I'm going to :bookmark: it for the next iteration of refactoring these methods ! Attempting this here is going to snowball into lots of finicky little edits to manually counting code comments etc in `getBOPSParams`.

src/export/bops/index.ts
```diff
@@ -0,0 +1,536 @@
...
+  passport: Passport;
+  sessionId: string;
+  flowName: string;
+}) {
+  const data = {} as BOPSFullPayload;
+  data.application_type = DEFAULT_APPLICATION_TYPE;
+
+  // Overwrite default application type if this isn't an LDC (relies on LDC flows having consistent slug)
+  //   eg because makeCsvData which is used across services calls this method
+  if (flowName && flowName !== "Apply for a lawful development certificate") {
```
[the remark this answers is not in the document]
jessicamcinchak: your right that `flowName` is required so this line was a bit redundant - all we really need to check here is what the flow name _is_: if it's LDC then use DEFAULT_APPLICATION_TYPE, otherwise set as flowName. 

supporting a value other than DEFAULT_APPLICATION_TYPE makes this function callable to generate CSV data that sharse the same data structure as BOPS but are used in non-LDC contexts.

src/export/bops/index.ts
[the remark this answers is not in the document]
DafyddLlyr: I'd also vote for removing `XXX` - it's new to me from John/Gunar also and I've always read it to mean "Care needed here / this is an explanation of something unexpected" which is just sort of what comments would be anyway.

src/export/bops/index.ts
DafyddLlyr: Thanks for the clarification here - a follow on PR makes perfect sense to me. 

This ticket is marked as blocked until we resolve this - https://trello.com/c/Cb5nFHIi/2297-99-share-section-names-with-bops

DafyddLlyr (approved):
Thank for the mammoth effort here and hidden complexities in porting this over 🏅 Also thanks @builditben for such a thorough review last week 👍 

Totally agree that we should...
 - Port this as is
 - Refactor and tidy up once it's here and we can test things in more isolation

src/flags.ts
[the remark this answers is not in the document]
jessicamcinchak: Yep this is a good shout, captured basic details in a Trello new request here please feel free to edit https://trello.com/c/f0SVmZbf/2452-store-result-flagset-values-in-a-database-table-and-expose-to-admin-editors

PR #23, opened 2023-04-27
build: don't recursively install dependencies

jessicamcinchak:


builditben:
The root cause of this issue looks like a missing dependency. I've patched that [here](https://github.com/theopensystemslab/planx-core/pull/24) which should fix the issue.

jessicamcinchak:
Closed in favor of #24

PR #24, opened 2023-04-27
chore: Add lodash dependencies

builditben:
An alternative solution to #23


jessicamcinchak (approved):
thanks for this ! 

to make sure i understand: we can reference specific lodash modules _only_ in dependencies, but need to keep a reference to the whole library @types still in dev dependencies which is what was previously missing?

builditben:
Yes, it looked like the `@types/lodash.isempty` and `@types/lodash.isnil` had an explicit dependency on `@types/lodash` which wasn't explicitly installed

PR #27, opened 2023-05-02
fix: Filter `templateNames` listed in `proposal.xml`

DafyddLlyr:
Comments for Daleel on failing Lambeth Uniform submissions - 

> the issue is that on the proposal.xml it was indicating that there are some attachments but they actually did not exist. for example, on porposal.xml it says there is an attachment called "LDCE.doc" but it actually does not exist.

This is because _all_ template names were being passed to the payload builder, not just the valid ones. I suspect this issue may be what's also affecting Southwark (minus the one timeout issue).

### DIFF
--- a/src/export/oneApp/index.test.ts
+++ b/src/export/oneApp/index.test.ts
@@ -1,0 +1,82 @@
+import { X2jOptionsOptional, XMLParser, XMLValidator } from "fast-xml-parser";
+import get from "lodash.get";
+import { OneAppPayload } from "./model";
+import { generateOneAppXML } from "./index";
+import { graphQLClient } from "../../graphql";
+import {
+  FileAttachment,
+} from "./types";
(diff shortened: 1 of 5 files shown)


src/export/oneApp/index.test.ts
```diff
@@ -0,0 +1,82 @@
...
+let mockHasRequiredDataForTemplate = jest.fn();
+jest.mock("@opensystemslab/planx-document-templates", () => {
+  return {
+    hasRequiredDataForTemplate: jest.fn(() => mockHasRequiredDataForTemplate()),
+  };
+});
+
+jest.mock("../../session", () => ({
+  getSessionById: jest.fn(() => mockSession)
+}));
```
jessicamcinchak, line 20: I was just wondering about options for query mocking over here in planx-core actually - this looks good & is helpful to see :+1:

PR #28, opened 2023-05-03
fix: Typo in `LookupHumanReadableProjectType` query

DafyddLlyr:


src/project-types.ts
```diff
@@ -19,7 +19,7 @@ async function lookupHumanReadableProjectTypes(
 ): Promise<string[]> {
   const query = gql`
     query LookupHumanReadableProjectType($rawList: [String!]) {
-      projectTypes: project_types(where: { value: { _in: $rawList } }) {
+      projectTypes: blpu_codes(where: { value: { _in: $rawList } }) {
```
jessicamcinchak, line 22: I'm a bit confused - `blpu_codes` are _property_ types not project types? Should the function & query names be updated here too?

DafyddLlyr: Apologies - I've clearly got a crossed wire here somewhere - it was failing due to the dummy data I was using, not this here 🤦

PR #29, opened 2023-05-11
fix: update get payment amount functionality

builditben:


builditben:
[Debug why frontend is failing to find Pay node and generate payment_request](https://trello.com/c/NMr8Rgl7/2449-debug-why-frontend-is-failing-to-find-pay-node-and-generate-paymentrequest)

src/logic.ts
```diff
@@ -94,3 +97,38 @@ export function sortBreadcrumbs(
...
+      node.parentId &&
+      lastCrumb.answers &&
+      lastCrumb.answers.includes(node.parentId!);
+    if (parentNodeIsCrumbAnswer && node.type === componentType) {
+      return node;
+    }
+  }
+  // when the above fails, fallback to a naïve scan of matching upcoming nodes
+  // FIXME: this approach does not respect flow branching logic
+  return truncatedFlow.find((n) => n.type === componentType);
```
builditben, line 133: This function is sufficient for invite-to-pay use cases (looking for the Send component which is always just after the last breadcrumb) but falls-back to an imperfect hack where the next node doesn't match. This may not work in every case so I'd recommend we use this function carefully until then.

builditben: I do have a plan for introducing more thorough node traversal logic in the near future (time permitting) so I will come back to this with a more robust approach.

src/types/flow.ts
```diff
@@ -19,10 +19,11 @@ export type FlowGraph = {
   [key: string]: Node;
 };
 
-export interface IndexedNode extends Node {
+export type IndexedNode = Node & {
   id: string;
+  type: ComponentType;
   parentId: string | null; // null if it is the first node
-}
+};
```
builditben, line 26: This is a `type` rather than an `interface` because it's a known well-defined structure and shouldn't have any additional undefined fields.

`IndexedNode`s should always have a defined type.

jessicamcinchak (approved):
Thanks for this! Haven't properly test driven in editor, but knowing this fixes the existing bug generating payment requests happy to merge before your end-of-week & test later/fix forward if necessary :slightly_smiling_face:

PR #30, opened 2023-05-15
feat: add CSV data as an export type

jessicamcinchak:
relies on BOPS work in #21 (will rebase this PR to main once that's merged, but this is most accurate diff)

generates the CSV data structure from a `sessionId` - then planx-new endpoints can handle the actual file-type/download logic as required in different in use cases.

in the future, I think our CSV data structure should aim to be more human-readable and not so closely coupled to the BOPS `proposal_details` format, but this is a v1 migration aiming for feature parity!


src/export/csv/index.ts
```diff
@@ -0,0 +1,101 @@
...
+      ],
+      metadata: { notes: file.applicant_description || "" },
+    });
+  });
+
+  // gather key reference fields, these will be first rows of CSV
+  const references: { question: string; responses: Array<Response> }[] = [
+    {
+      question: "Planning Application Reference", // match language used on Confirmation page
+      responses: [{ value: sessionId }],
```
jessicamcinchak, line 61: I updated these references from the original implementation to adhere to the `responses` type, but realized that wrapping in `value` does break the HTML overview parsing in a couple places which I haven't updated yet in planx-document-templates.

To sense check: does having a consistent `responses` type always wrapped in value seem right here, or is `responses: string | Array<Response>` not a problem?

builditben (approved):

src/export/csv/index.ts
```diff
@@ -0,0 +1,101 @@
...
+import { getSessionById } from "../../session";
+import { getBOPSParams } from "../bops/index";
+import { QuestionAndResponses, Response } from "../bops/model";
+import { CSVData } from "./model";
+
+export async function generateCSVData(
+  client: GraphQLClient,
+  sessionId: string
+): Promise<CSVData> {
+  const session = await getSessionById(client, sessionId);
```
DafyddLlyr, line 13: This may be in another PR, but I believe using `generateCSVData()` will only work via the `_admin` client due to missing permissions - we won't be able to use this as-is in the frontend.

We can address this in a number of ways but it's probably something pretty fundamental to how this library works that merits a discussion. I've added it to the dev call agenda 👍

src/export/csv/index.ts
```diff
@@ -0,0 +1,101 @@
...
+        {
+          value: [
+            bopsData.site?.address_1,
+            bopsData.site?.address_2,
+            bopsData.site?.town,
+            bopsData.site?.postcode,
+          ]
+            .filter(Boolean)
+            .join(" ")
+            .replaceAll(",", ""), // omit commas for csv > pdf parsing later by Uniform
```
DafyddLlyr, line 75: This is for custom solutions right, such as Kev's CSV -> PDF generation for Bucks?

If so, won't any string with commas break this generation, e.g. a project description of "something, something, and something".

It might be better to parse the entire string, escaping strings containing commas with double quotes. There may well be an option for this in the CSV library we're using?

Appreciate this isn't a new issue introduced here but worth flagging and we can raise an issue if this is going to be fixed later 👍

DafyddLlyr (approved):

src/export/csv/index.ts
jessicamcinchak: This shout about Kev's custom solution is a good one & I just remembered this also while going to update `planx-document-templates`. 

I think it's actually a reason _not_ to change the CSV formatting yet to align all rows to the `Response` type - just because I don't think we have any confirmation that Majad has taken over that workflow and it might result in a slow testing turnaround time / even more processing woes. 

Let's come back to the improving the actual structure of the CSV and removing parsing hacks like this line in a separate PR :+1: I'll keep feature parity with the current CSV for now.

PR #31, opened 2023-05-22
feat: Add authorizeSession to graphql client

builditben:
This method allows a public client to access data for a given session by amending the GraphQL headers of the client.

### DIFF
--- a/src/graphql.ts
+++ b/src/graphql.ts
@@ -3,11 +3,23 @@ import { GraphQLClient } from "graphql-request";
 export function graphQLClient({
   url,
   secret,
+  authorizedSession,
 }: {
   url: string;
-  secret?: string | undefined;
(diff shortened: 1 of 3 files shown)


DafyddLlyr (approved):
This is great! Nice and simple 👏 

I agree that the actual headers (`x-hasura-lowcal-session-id`) should be the responsibility of this library and just handing in an `authorizedSession` is a nice clean API.

Going forward it would be good to think about how we could handle different authorisation types - not all requests will require an email + session id. For example, fetching a payment request from `public` requires `x-hasura-payment-request-id` but nothing else.

I think it's easy to expand on this solution to accommodate things like this but it might be worth taking stock and considering our options before we do that - I've added something to the agenda for tomorrow's dev call 👍

PR #33, opened 2023-05-24
chore: Tidy-ups and package upgrades

builditben:
A few small fixes and updates that address linting and maintainability issues.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,5 +1,6 @@
 name: Test
 on:
+  workflow_dispatch:
   pull_request:
     types:
       - opened
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,6 +29,9 @@ jobs:
           version: 7
           run_install: true
 
+      - name: Clean generated assets
+        run: rm -rf ./dist ./types
+
       - name: Run code checks
         run: pnpm check
 
--- a/src/export/bops/tests/payload.test.ts
+++ b/src/export/bops/tests/payload.test.ts
@@ -14,12 +14,14 @@ describe("Full BOPS payload", () => {
 
   it("builds the payload as expected", () => {
     // check the response shape excluding proposal_details as order can vary
-    const generatedMinimumPayload = (({ proposal_details, ...p }) => p)(
-      generatedPayload
-    );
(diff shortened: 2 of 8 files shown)


.github/workflows/test.yml
```diff
@@ -1,5 +1,6 @@
 name: Test
 on:
+  workflow_dispatch:
```
builditben, line 3: Checks were failing on the main branch for me locally (??). I've added the ability to manually trigger tests to be able to sense check things like this.

src/export/bops/tests/payload.test.ts
```diff
@@ -14,12 +14,14 @@ describe("Full BOPS payload", () => {
 
   it("builds the payload as expected", () => {
     // check the response shape excluding proposal_details as order can vary
-    const generatedMinimumPayload = (({ proposal_details, ...p }) => p)(
-      generatedPayload
-    );
-    const mockExpectedMinimumPayload = (({ proposal_details, ...p }) => p)(
-      mockExpectedBOPSPayload
-    );
```
builditben, line 22: This is clever but simple and explicit is preferable.

src/models/flags.ts
```diff
@@ -68,32 +68,37 @@ export const categoriesAndFlags: Record<string, Record<string, string[]>> = {
   },
 };
 
-export const flatFlags: Array<Flag> = [];
```
builditben, line 71: This looks like we're exporting an empty array but we're actually setting up an array that is mutated later.
This is a lot easier to understand if we export *after* we've constructed the array.

src/models/session/logic.test.ts
```diff
@@ -9,12 +9,12 @@ import * as simple from "./mocks/simple-flow-breadcrumbs";
...
 import * as large from "./mocks/large-real-life-flow";
 import * as branching from "./mocks/branching-flow";
-import { ComponentType } from "../../../types";
+import { ComponentType } from "../../types";
 import type {
   OrderedFlow,
   NormalizedFlow,
   OrderedBreadcrumbs,
-} from "../../../types";
+} from "../../types";
```
builditben, line 17: I'm still a bit puzzled about how these reference errors slipped through.

PR #34, opened 2023-05-24
feat: Add expand flow function to replace normalize flow

builditben:
This is more groundwork for a full set of logic functions outside of the editor.

The idea here is to build, as a data structure, the shape of a flow to allow easy traversal. I've not written many actual logic functions yet, but, to avoid a mega-large PR, I'm submitting the initial data structure/types work.


src/models/session/index.ts
```diff
@@ -1,2 +1 @@
 export * from "./logic";
-export * from "./session-state";
```
builditben, line 2: SessionState is still experimental at this stage so I'm not exporting it yet.

src/models/session/logic.test.ts
```diff
@@ -112,9 +130,10 @@ describe("findNextNodeOfType", () => {
     });
     expect(nextNode).toEqual({
       id: "secondQuestion",
+      rootNodeId: "secondQuestion",
       parentId: "firstQuestion",
       type: ComponentType.Question,
-      edges: ["secondAnswer"],
+      edges: ["Q2Answer1", "Q2Answer2", "Q2Answer3"],
```
builditben, line 136: The test flow has been updated to include more branches (no functional change here).

src/models/session/mocks/complex-flow-breadcrumbs.ts
```diff
@@ -600,96 +613,93 @@ export const normalizedFlow: NormalizedFlow = [
...
+                  color: "#EFEFEF",
+                  title: "?, so must be a 🍌 or 🔧",
+                  resetButton: false,
+                },
+                type: ComponentType.Notice,
+              },
+            ],
+          },
+        },
+      ],
```
builditben, line 701: Hopefully, this data structure is intuitive and self-explanatory. You can see how the structure of the flow is encoded as a data structure which could easily be traversed.

builditben, line 701: Hopefully, this data structure is intuitive and self-explanatory. You can see how the structure of the flow is encoded as a data structure which could easily be traversed.

src/types/component.ts
```diff
@@ -31,3 +33,230 @@ export enum ComponentType {
...
+  properties?: {
+    reads?: readonly PropertyDefinition[];
+    writes?: readonly PropertyDefinition[];
+  };
+  branching?: BranchingRule;
+  placementRestriction?: {
+    previous?: readonly ComponentType[]; // previous node must be one of these types
+    next?: readonly ComponentType[]; // next node must be one of these types
+  };
+};
```
builditben, line 54: I'm letting the types drive the design here a little (i.e. some of these fields aren't used yet). 
Here, the functionality of components are declared as a data structure. This means that logic functions can query the intended behaviour of the component explicitly. 
For example, each component has branching rules which define what nodes can be nested and what kind of collection the nesting represents. This allows for traversal without having to special case particular component types in the logic functions.

src/types/component.ts
```diff
@@ -31,3 +33,230 @@ export enum ComponentType {
...
+    },
+  },
+  PropertyInformation: {
+    type: ComponentType.PropertyInformation,
+  },
+
+  // Defunct(?)
+  Flow: {
+    type: ComponentType.Flow,
+  },
```
builditben, line 261: Is this type used at all?

src/types/flow.ts
```diff
@@ -19,18 +19,27 @@ export type FlowGraph = {
   [key: string]: Node;
 };
 
-export type IndexedNode = Node & {
+export type IndexedNode = {
   id: string;
   type: ComponentType;
   parentId: string | null; // null if it is the first node
+  rootNodeId: string;
```
builditben, line 26: It's helpful for `IndexedNode`s to pass on information about where in the `_root.edges` that node is found/nested. This means that a flat list of `IndexedNode`s can be used to find a more meaningful `StructuredNode` - both of these data structures working in tandem make it easy to perform all kinds of traversals and look-ups.

src/models/session/index.ts
DafyddLlyr: I'm still working my way through this PR but wanted to flag early that this (merged) PR is importing `StaticSessionState` https://github.com/theopensystemslab/planx-core/pull/35/files

This hasn't gone to prod yet, and I haven't yet informed BOPS that section names have been added to the payload so it's fine if we need to revert this and find another solution for this one - just let me know.

builditben: I'll submit a patch to deprecate `StaticSessionState` usages in planx-new before merging this if that's okay. That means we can play around a little more with the interface and structure here without impacting anything for the time-being. I hope to get this in a production ready state soon though.

src/models/session/session-state.ts
```diff
@@ -1,187 +1,110 @@
...
+    return this.orderedFlow.reverse().find((node) => node.id === id);
   }
 
-  private currentIndex(breadcrumbs: OrderedBreadcrumbs) {
-    const lastCrumb: NormalizedCrumb = breadcrumbs.at(-1)!;
-    const currentIndex = this.normalizedFlow.findIndex((node) => {
-      if (lastCrumb.answers?.length) {
-        const lastAnswerId = lastCrumb.answers.at(-1);
-        return node.id === lastAnswerId;
+  private findStructuredNode(
```
DafyddLlyr, line 86: nit: This function is clearly pretty important and called for multiple operations. I think the naming here is clear and meaningful but I'd personally be inclined to add a few comments just to be super explicit about what each step is doing with long-term maintenance in mind.

DafyddLlyr (approved):
A lot of good work here, and a huge amount of design and thought 🙌 It pulls together a lot of our previous discussions and half-discussions pretty neatly.

We'll need to consider the next steps carefully given the time we have but this is such a solid start that I'm excited to see more of.

src/models/session/session-state.ts
DafyddLlyr: I appreciate this a bit contradictory sorry and we shouldn't need comments when it's clear and meaningful, but I guess I'm thinking of future-proofing here as best we can.

PR #36, opened 2023-05-26
chore: Temporary skip of payment validation

DafyddLlyr:


builditben:
@DafyddLlyr I think we can close this now. Is that right?

DafyddLlyr:
Thanks for flagging @builditben - PR closed!

PR #39, opened 2023-05-31
feat: Add namespaced clients

builditben:
The purpose of this PR is to introduce some new request functions used in e2e testing.

Rather than adding to an ever-growing set of exported request functions in one file, here I'm introducing namespaced clients. Each client exposes a nice simple interface within a namespace of the main client (e.g. `client.session.lock("123")`).

Used by https://github.com/theopensystemslab/planx-new/pull/1743

### DIFF
--- a/src/requests/application.ts
+++ b/src/requests/application.ts
@@ -112,14 +112,14 @@ export async function _destroyBopsApplication(
     await client.request(
       gql`
         mutation DestroyBopsApplication($sessionId: String!) {
-          delete_bops_applications_by_pk(id: $sessionId) {
-            id
+          delete_bops_applications(where: { session_id: { _eq: $sessionId } }) {
+            affected_rows
           }
         }
       `,
       { sessionId }
     );
-  return Boolean(response.delete_bops_applications_by_pk?.id);
+  return Boolean(response.delete_bops_applications?.affected_rows);
 }
 
 export async function _destroyEmailApplication(
@@ -130,14 +130,16 @@ export async function _destroyEmailApplication(
     await client.request(
       gql`
         mutation DestroyEmailApplication($sessionId: uuid!) {
-          delete_email_applications_by_pk(id: $sessionId) {
-            id
+          delete_email_applications(
+            where: { session_id: { _eq: $sessionId } }
+          ) {
+            affected_rows
           }
         }
       `,
       { sessionId }
     );
-  return Boolean(response.delete_email_applications_by_pk?.id);
+  return Boolean(response.delete_email_applications?.affected_rows);
 }
 
 export async function _destroyUniformApplication(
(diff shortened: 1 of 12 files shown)


src/requests/payment-request.ts
```diff
@@ -3,10 +3,35 @@ import { getDetailedSessionById } from "./session";
...
+    payeeName: string;
+    payeeEmail: string;
+    sessionPreviewKeys: Array<KeyPath>;
+  }): Promise<PaymentRequest> {
+    return createPaymentRequest(this.client, args);
+  }
+
+  async _markAsPaid(paymentRequestId: string): Promise<boolean> {
+    return _markPaymentRequestAsPaid(this.client, paymentRequestId);
+  }
```
builditben, line 28: I'm introducing a pattern here to prefix methods which shouldn't normally be called. These are methods used by tests to manipulate state. The prefix is a strong signal that this shouldn't be used in normal business logic.

This convention does (again) clash with the convention of unused named variables and it's also a common convention for private variables but I think it works here since it's always part of a chained method call rather than a bare variable by itself. Feel free to suggest alternatives if you disagree.

src/requests/index.ts
```diff
@@ -21,18 +27,38 @@ import type { GraphQLClient } from "graphql-request";
 const defaultURL = process.env.HASURA_GRAPHQL_URL;
 
 export class CoreDomainClient {
-  client: GraphQLClient;
+  client!: GraphQLClient;
```
builditben, line 30: the `!` lets TS know that I'm handling setting this attribute outside of the constructor (in this case, in `setClient`.

src/requests/application.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  async uniformResponse(
+    sessionId: string
+  ): Promise<ApplicationResponse | null> {
+    return uniformApplicationResponse(this.client, sessionId);
+  }
+}
+
+export async function uniformApplicationResponse(
```
builditben, line 26: I'm exporting `ApplicationClient` and also still exporting the raw functions for two reasons:
1. this pattern is consistent with what we already have
2. when we introduce a test suite for these request functions, this pattern should make mocking easier (depending on the approach)

src/requests/index.ts
```diff
@@ -41,12 +67,18 @@ export class CoreDomainClient {
...
       url: this.url,
       authorizedSession: sessionDetails,
     });
+    this.setClient(client);
   }
 
+  // TODO: refactor below into client namespaces (e.g. SessionClient)
+  // namspacing prevents this class from growing too unwieldy while still allowing for
+  // a simple interface for callers.
+  // Namespacing also allows for terser function names (e.g. `client.session.lock("123")`)
```
builditben, line 80: I suggest the new pattern of exporting a namespaced client means that we deprecate the existing pattern and stop adding to it.

DafyddLlyr (approved):
This is a nice and elegant solution to the growing index file, and is a nice mental model to work from 👌

PR #40, opened 2023-06-08
chore: refactor export functions to improve performance

builditben:
This PR intends to improve the performance of the export functions. I'm not sure that it will have as big an impact as I had hoped.

It turns out that I wasn't correct in my assumptions about how Javescript is passing values and it's memory model. JS uses a [call by sharing](https://en.wikipedia.org/wiki/Evaluation_strategy#Call_by_sharing) evaluation strategy which means that objects can be passed around without copying the object itself.

I initially refactored to introduce fetchers in the compute functions but, after learning that this strategy will not impact performance, I reverted to a more general refactor to see where else I could make improvements.

`StaticSessionState` was creating a new normalized flow on construction. I've replaced this with `sortFlow` which should have similar performance characteristics so I'm not expecting a huge performance difference here either. Once we have access to a sorted flow from a db query - the compute time and memory usage here should go down significantly.

So this PR does a little tidying-up and there are a few changes which I think could improve performance a small amount but I don't think this will actually fix the issues with processing very large flows. :disappointed:


src/requests/export.ts
```diff
@@ -0,0 +1,69 @@
...
+}
+
+export async function generateBOPSPayload(
+  client: GraphQLClient,
+  sessionId: string
+): Promise<BOPSFullPayload> {
+  try {
+    const session = await getSessionById(client, sessionId);
+    if (!session) throw new Error(`Cannot find session ${sessionId}`);
+    const flow = await findPublisedFlowBySessionId(client, sessionId);
```
builditben, line 55: This is a functional change which I think we want. `findPublishedFlowBySessionId` is looking up the flow and calling `getMostRecentPublishedFlowBeforeTimestamp` to get the flow as it was published when the session was last updated. It should mean that the export can happen at any time but always points to the flow as it was when the session was completed.

src/requests/session.ts
```diff
@@ -14,11 +15,11 @@ export class SessionClient {
     this.client = client;
   }
 
-  async find(sessionId: string): Promise<Session> {
+  async find(sessionId: string): Promise<Session | null> {
```
builditben, line 18: I'm fixing some of these type defs where Hasura can return `null` when the resource isn't found.

src/types/result.ts
```diff
@@ -2,10 +2,7 @@ export type resultOverrides = {
   [flagId: string]: { heading?: string; description?: string };
 };
 
-export type ResultData = (
-  flagSet?: string,
-  overrides?: resultOverrides
-) => {
+export type ResultData = {
```
builditben, line 5: This type is less complex to work with if it references the output directly rather than having to reference the more indirect `ReturnType<ResultData>`

src/requests/session.ts
```diff
@@ -104,6 +105,44 @@ export async function getDetailedSessionById(
...
+      query GetSessionById($id: uuid!) {
+        lowcal_sessions_by_pk(id: $id) {
+          breadcrumbs: data(path: "breadcrumbs")
+        }
+      }
+    `,
+    { id: sessionId }
+  );
+  return response?.lowcal_sessions_by_pk?.breadcrumbs || null;
+}
```
builditben, line 126: This function was used in earlier versions of this refactor but not anymore. I thought it was probably worth keeping for future reuse.

src/export/bops/index.ts
```diff
@@ -110,52 +85,55 @@ function exhaustiveCheck(type: never): never {
...
+    id: string,
+    metadata: QuestionMetaData
+  ): QuestionMetaData => {
+    const currentNode = sortedFlow.find((node) => node.id === id);
+    const currentSection = sortedFlow.find(
+      (node) => node.id == currentNode!.sectionId
+    );
+    metadata.section_name = currentSection?.data?.title as string;
+    return metadata;
+  };
```
builditben, line 109: I've inlined these closures based on my initial hunch that this could improve performance but I'm not sure that's actually the case.

src/export/bops/tests/payload.test.ts
```diff
@@ -14,14 +15,12 @@ describe("Full BOPS payload", () => {
...
-    const mockExpectedMinimumPayload = {
-      ...mockExpectedBOPSPayload,
-      proposal_details: undefined,
-    };
+    const generatedMinimumPayload = omit(generatedPayload, [
+      "proposal_details",
+    ]);
+    const mockExpectedMinimumPayload = omit(mockExpectedBOPSPayload, [
+      "proposal_details",
+    ]);
```
DafyddLlyr, line 23: Much cleaner and imho more meaningful

src/requests/export.ts
```diff
@@ -0,0 +1,69 @@
+import { GraphQLClient } from "graphql-request";
+import { getSessionById, getSessionPassport } from "./session";
+import { findPublisedFlowBySessionId, getFlowName } from "./flow";
+import { computeCSVData } from "../export/csv";
+import { computeBOPSParams } from "../export/bops";
+import type { CSVData } from "../export/csv/model";
+import type { BOPSFullPayload } from "../export/bops/model";
+
+export class ExportClient {
```
DafyddLlyr, line 9: Would it make sense to bring OneApp into here also?

src/requests/export.ts
DafyddLlyr: This is a good catch and I agree is correct - without this we could have different sections, different passport vars, etc which would impact the representation of the user's journey.

src/models/result.ts
```diff
@@ -1,24 +1,29 @@
...
-export const getResultData = (
-  breadcrumbs: Breadcrumbs,
-  flow: LooseFlowGraph,
-  flagSet: Parameters<ResultData>[0] = DEFAULT_FLAG_CATEGORY,
-  overrides?: resultOverrides
-) => {
+export function computeResultData({
+  flow,
+  breadcrumbs,
+  flagSet = DEFAULT_FLAG_CATEGORY,
```
DafyddLlyr, line 19: 👍

src/types/result.ts
DafyddLlyr: I agree with the simplification here - having a complex type and then referencing both it's `ReturnType<>` and `Parameters<>` is clever, but not necessarily that easy to follow.

DafyddLlyr (approved):
This is great 👍 Thanks for picking this up and, as always, tidying up as you go - always good to see 😄

src/export/bops/index.ts
```diff
@@ -110,179 +86,191 @@ function exhaustiveCheck(type: never): never {
...
+      .concat([{ id: "_root", name: "_root", edges: flow._root.edges }])
+      .reverse();
+
+  const portalName = (id: string): string | undefined => {
+    if (id === "_root") return "_root";
+    for (const portal of portals) {
+      if (portal.edges?.includes(id)) {
+        return portal.name;
+      }
+    }
```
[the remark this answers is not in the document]
jessicamcinchak: Now that all LDC services are using sections, I think it's also worth asking BOPS if we can retire calculating/sending the portal name all together here - portal name was always intended to be a temporary proxy until sections existed anyways. Already planning to talk to them about constraints next Tuesday, happy to bring this up too.

DafyddLlyr (approved):

PR #42, opened 2023-06-13
refactor: Fetch flags from the DB

builditben:


src/export/bops/index.ts
```diff
@@ -38,13 +40,17 @@ export async function generateBOPSPayload(
       `Cannot get flow ${session.flowId}, therefore cannot generate BOPS payload.`
     );
   const flowName = await getFlowName(client, session.flowId);
+  const flags = await getAllFlags(client);
+  const resultFlagSet = await getFlagSet(client, "PlanningPermission");
```
builditben, line 44: Here we're still hard-coding which flag set to use for result flags.

Ideally, we'd find the result component in breadcrumb (assuming there is only one) and see which result flag set is set there. I think this makes the transition towards something like that fairly straight-forward.

PR #52, opened 2023-07-04
chore: Update GovPay types

DafyddLlyr:
As part of the work outlined [here](https://github.com/theopensystemslab/planx-new/pull/1902) I'm pulling in types from a single source - they're currently replicated in planx-core, the API, and the Editor.


jessicamcinchak (approved):
sorry I missed this dependency first time around when approving the planx-new PR !

PR #53, opened 2023-07-05
feat: Add next steps to component types list

ianjon3s:
PR adds the NextSteps component to component types list, in sync with planx-new.

### DIFF
--- a/src/types/component.ts
+++ b/src/types/component.ts
@@ -30,4 +30,5 @@ export enum ComponentType {
   Send = 650,
   Calculate = 700,
   Confirmation = 725,
+  NextSteps = 730,
 }


DafyddLlyr:
We also need to update and add this component in `src/export/bops/index.ts` to return `false` as it's not a an input which the user can use for submission info (e.g. like a text field or address).

ianjon3s:
> We also need to update and add this component in `src/export/bops/index.ts` to return `false` as it's not a an input which the user can use for submission info (e.g. like a text field or address).

Thanks Daf, I've updated.

DafyddLlyr (approved):

PR #55, opened 2023-07-10
chore: Merge document templates

ghost:
This PR merges functionality from `@theopensystemslab/planx-document-templates` with some minor interface changes.

This should allow us to deprecate the planx-document-templates module.

Testing:
- this is implemented in https://github.com/theopensystemslab/planx-new/pull/1951
- a set of example documents should be generated and attached to the summary of each test run: https://github.com/theopensystemslab/planx-core/actions/runs/5518040608


.github/workflows/test.yml
```diff
@@ -37,3 +37,13 @@ jobs:
...
+
+      - name: Build & generate examples
+        run: pnpm examples
+
+      - name: Save examples
+        uses: actions/upload-artifact@v3
+        with:
+          name: examples
+          path: examples
+          retention-days: 3
```
DafyddLlyr, line 49: This is a nice way of testing this 👍

DafyddLlyr (approved):
Tested locally and all working as expected - tests passing and documents generating ✅

PR #68, opened 2023-07-18
feat: update session client and add integration test suite

builditben:
This PR updates session requests to write/read to/from the `sessions` table and the `lowcal_sessions` table with a consistent interface. Integration tests are also added in this PR.

The `OrderedSession` type returned from session lookups (e.g. `$admin.session.find(id)` ) contains enough information about the flow to allow subsequent processing to not require flow look-ups. This should save significant memory resources during submission processing.

This PR depends on a few tidy-ups in #33


package.json
```diff
@@ -33,6 +33,7 @@
     "lint:fix": "eslint --fix 'src/**/*.{js,ts}' && prettier -w src/**/*",
     "test": "jest",
     "test:coverage": "jest; open ./coverage/lcov-report/index.html",
+    "test:integration": "cd tests && pnpm it",
```
DafyddLlyr, line 36: `pnpm it` doesn't seem to correlate to a script in `/tests/package.json`

scripts/setup-integration-test-containers.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+# clone the main planx repo
+git clone -b ben/add-payment-id-computed-field https://github.com/theopensystemslab/planx-new planx
```
DafyddLlyr, line 4:
```suggestion
# Change to point towards a specific branch
BRANCH="main"
git clone -b $BRANCH https://github.com/theopensystemslab/planx-new planx
```

src/models/session/logic.ts
```diff
@@ -94,6 +96,56 @@ export function sortBreadcrumbs(
...
+      answers: crumb.answers,
+      data: crumb.data,
+      override: crumb.override,
+      feedback: crumb.feedback,
+    };
+    return acc;
+  }, {} as Breadcrumbs);
+}
+
+export function computePassport(breadcrumbs: OrderedBreadcrumbs): IPassport {
```
DafyddLlyr, line 114: TODO: Test this using existing `computePassport()` tests in - 
 - `planx-new/editor.planx.uk/src/pages/FlowEditor/lib/__tests__/preview.test.ts`
 - `planx-new/editor.planx.uk/src/pages/FlowEditor/lib/__tests__/overrideAnswer.test.ts`
 - `planx-new/editor.planx.uk/src/pages/FlowEditor/lib/__tests__/granularity.test.ts`

src/requests/index.ts
```diff
@@ -98,3 +105,56 @@ export class DomainClient {
     return formatRawProjectTypes(this.client, rawProjectTypes);
   }
 }
+
+// test set-up/tear-down utility
+export class TestClient {
+  private url: string;
+  private secret: string;
+
+  constructor({
```
DafyddLlyr, line 114: Could we throw an error here if `NODE_ENV !== "test"`? 

It's imperfect, but could act as somewhat of a guard against improper use and would make this more self-documenting at a minimum.

https://jestjs.io/docs/environment-variables#node_env

src/requests/session.ts
```diff
@@ -16,151 +25,210 @@ export class SessionClient {
...
-    flowId: string;
-    data: Partial<SessionData>;
+    flowId: FlowId;
+    breadcrumbs: OrderedBreadcrumbs;
     email?: string;
-  }) {
-    return createSession({ client: this.client, flowId, data, email });
+  }): Promise<SessionId> {
+    const sessionId: SessionId = uuidV4();
+    await Promise.all([
```
DafyddLlyr, line 38: It may be slightly more complex / less readable, but I think we'd be better treating these CRUD operations on `sessions` and `lowcal_sessions` as single transaction (i.e. two mutations within the same query).

src/requests/session.ts
```diff
@@ -16,151 +25,210 @@ export class SessionClient {
...
+        sessionId,
+        flowId,
+        breadcrumbs,
+        email,
+      }),
+    ]);
+    return sessionId;
+  }
+
+  async find(sessionId: SessionId): Promise<OrderedSession | null> {
```
DafyddLlyr, line 57: 👍

tests/src/session.test.ts
```diff
@@ -0,0 +1,321 @@
...
+  getFlow,
+  getLowcalSession,
+  getSession,
+  setUpContext,
+  tearDownTestContext,
+} from "./context";
+
+const sessions: SessionScenarios = {};
+
+beforeAll(async () => {
```
DafyddLlyr, line 21: Tests here are super clear and helpful.

PR #78, opened 2023-07-26
chore: Tighten types

builditben:
This PR tunes some of the types, removes loose types and resolves all outstanding linting errors. It also re-introduces a `tsc` check for `implicitAny`.

This PR makes a few changes which _could_ have changed the output types and I was a little unsure of some of the export typings in particular. A thorough review from more seasoned devs would be very much appreciated!


src/export/bops/index.ts
```diff
@@ -1,30 +1,33 @@
...
+  BOPSFullPayload,
+  ComponentType,
+  DEFAULT_APPLICATION_TYPE,
+  flatFlags,
+  GOV_PAY_PASSPORT_KEY,
   USER_ROLES,
 } from "../../types";
 
-const bopsDictionary = {
+const bopsDictionary: { [Property in keyof BOPSFullPayload]?: string } = {
```
DafyddLlyr, line 30: A better/tighter approach here might just be to use `as const` for this variable.

Instead of casting to `string` on line 145 we can then just use a type guard instead. Something like -

```ts
const isValidKey = (fn: unknown): fn is keyof typeof bopsDictionary => !Object.values(bopsDictionary).find((val) => val === fn)
```

src/export/bops/index.ts
```diff
@@ -267,30 +282,29 @@ export function computeBOPSParams({
...
 
   // 1b. property boundary
-  const geojson = passport.data?.["property.boundary.site"];
+  const geojson = passport.any(["property.boundary.site"]);
   if (geojson) data.boundary_geojson = geojson;
 
   // 2. files
   Object.entries(passport.data || {})
-    .filter(([, v]: any) => v?.[0]?.url)
+    .filter(([, v]) => (v as { url: string }[] | undefined)?.[0]?.url)
```
DafyddLlyr, line 305: It looks like we could use `Passport.files()` here

src/export/csv/index.ts
```diff
@@ -37,21 +38,21 @@ export function computeQuestionAndResponses({
...
   // format dedicated BOPS properties (eg `applicant_email`) as list of questions & responses to match proposal_details
   //   omitting debug data and payload keys already in confirmation details
-  const summary: Record<string, any> = {
+  const summary: Partial<BOPSFullPayload> = {
     ...omit(bopsData, ["planx_debug_data", "files", "proposal_details"]),
   };
   const formattedSummary: {
     question: string;
-    responses: Array<Response>;
+    responses: Response[];
```
DafyddLlyr, line 50: 👍

src/types/bops.ts
```diff
@@ -33,10 +12,10 @@ interface BOPSMinimumPayload {
...
     town?: string;
     postcode?: string;
-    latitude: string;
-    longitude: string;
-    x: string;
-    y: string;
+    latitude: number;
+    longitude: number;
+    x: number;
+    y: number;
```
DafyddLlyr, line 18: Follow on from dev call - these are all currently sent to BoPS as `number` so this is correct. I checked a few recent records on `bops_applications`.

src/types/bops.ts
```diff
@@ -82,6 +60,8 @@ export interface BOPSFullPayload extends BOPSMinimumPayload {
 
 export const USER_ROLES = ["applicant", "agent", "proxy"] as const;
 
+export type UserRole = (typeof USER_ROLES)[number];
```
DafyddLlyr, line 63: 👍

DafyddLlyr (approved):
A few comments here but I'm not expecting you to pick them up - more a list of thing for me and/or Jess to resolve.

src/models/passport/passport.ts
```diff
@@ -62,6 +62,12 @@ export class Passport {
     if (!this.has(path)) return;
     return get({ data: this.data, path })!;
   }
+
+  generic<T>(path: string): T | undefined {
```
DafyddLlyr, line 66: I'm planning another pass at this class and have just introduced this here to solve a few type problems.

I think this interface is essentially what I'd like a `Passport.get()` method to look like (but the internal logic would more closely map the current `get()` function.

src/export/digitalPlanning/model.ts
```diff
@@ -420,14 +438,15 @@ export class DigitalPlanning {
 
   private getApplicationFee(): Payload["data"]["application"]["fee"] {
     const baseFee = {
-      calculated: this.passport.data?.["application.fee.calculated"],
-      payable: this.passport.data?.["application.fee.payable"],
+      calculated:
+        this.passport.generic<number>("application.fee.calculated") || 0,
```
DafyddLlyr, line 442: `genric` is very much a temporary placeholder name, but it allowed me to make progress and test this without changing the current `get()` functionality in an already large PR.

src/export/digitalPlanning/model.ts
```diff
@@ -495,21 +511,22 @@ export class DigitalPlanning {
...
+        } as ProjectType;
+      }),
       description:
         this.passport.data?.["proposal.description"] || "Not provided",
+      boundary: this.getBoundary(),
       date: {
-        start: this.passport.data?.["proposal.start.date"],
-        completion: this.passport.data?.["proposal.completion.date"], // this.passport.data?.["proposal.finish.date"],
+        start: this.passport.generic<string>("proposal.start.date")!,
+        completion: this.passport.generic<string>("proposal.completion.date")!, // this.passport.data?.["proposal.finish.date"],
```
jessicamcinchak, line 529: I'm a little fuzzy on this TS syntax - are the `!` indicating required/always expected here? start & completion dates are optional per the schema, and are not expected to be provided throughout

jessicamcinchak (approved):
Looked closely at `digitalPlanning` model, but trusting tests for other areas ! 

All this seems okay to me, pending one small comment about start/completion dates :+1:

src/export/digitalPlanning/model.ts
DafyddLlyr: Good spot! This may have been something which was required on my first pass through here.

DafyddLlyr:
> Looked closely at `digitalPlanning` model, but trusting tests for other areas !

Yep - tests are doing their job here! Despite a fair number of changes, no tests needed to be updated which is a good sign.

PR #91, opened 2023-08-08
chore(types): Make `Passport.data` required property

DafyddLlyr:
Required for https://github.com/theopensystemslab/planx-new/pull/2074


jessicamcinchak (approved):
Makes sense !

PR #97, opened 2023-08-15
feat: Add team boundary argument to map and html creation

DafyddLlyr:
## What does this PR do?
 - Requires that an extent be passed to map and HTML templates
 - Uses Bucks as an example extent GeoJSON (see attached generated examples)
 - Re-enables map pan and zoom

📂  [Examples.zip](https://github.com/theopensystemslab/planx-core/files/12352683/Archive.zip)

Currently live on this environment - https://github.com/theopensystemslab/planx-new/pull/2099

### DIFF
--- a/src/templates/html/application/ApplicationHTML.tsx
+++ b/src/templates/html/application/ApplicationHTML.tsx
@@ -290,6 +290,7 @@ export function ApplicationHTML(props: {
                     showNorthArrow={true}
                     showScale={true}
                     useScalebarStyle={true}
+                    hideResetControl={true}
                     geojsonData={JSON.stringify(boundary)}
                     id="boundary-map"
                     showPrint={true}
--- a/src/templates/html/application/ApplicationHTML.tsx
+++ b/src/templates/html/application/ApplicationHTML.tsx
@@ -1,16 +1,17 @@
 import { css, Global } from "@emotion/react";
 import Box from "@mui/material/Box";
 import Grid from "@mui/material/Grid";
+import groupBy from "lodash.groupby";
 import prettyTitle from "lodash.startcase";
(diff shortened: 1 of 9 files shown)


jessicamcinchak:
As discussed, this should now be able to call `passport.data.["property.boundary.site.buffered"]` for simplicity and a more restricted extent (75m) based on changes here https://github.com/theopensystemslab/planx-new/pull/2162 !

src/templates/html/application/ApplicationHTML.tsx
```diff
@@ -283,14 +287,13 @@ export function ApplicationHTML(props: { data: PlanXExportData[] }) {
               {boundary && (
                 <Box sx={{ marginBottom: 1 }}>
                   <my-map
-                    staticMode={true}
                     showNorthArrow={true}
                     showScale={true}
                     useScalebarStyle={true}
-                    hideResetControl={true}
```
jessicamcinchak, line 290: so we actually still want `hideResetControl={true}` here I think as this only hides the reset button / the geojson should always be static on an overview document unlike an active drawMode map! We just want to remove `staticMode` which then displays the zoom controls again.

src/templates/html/map/Map.tsx
```diff
@@ -10,24 +10,27 @@ declare global {
...
+export default function Map(props: {
+  boundary: object;
+  clipGeojsonData: GeoJSON.Feature;
+}) {
   return (
     <my-map
-      staticMode={true}
       showNorthArrow={true}
       showScale={true}
-      hideResetControl={true}
```
jessicamcinchak, line 29: same as above

jessicamcinchak (approved):

PR #100, opened 2023-08-18
feat: Initial setup of new data export format

DafyddLlyr:
## What does this PR do?

 - Adds a new data export format - DigitalPlanning (terrible name, let's please change this soon!)
 - Re-generates types from JSON schema document
 - It's just using dummy data for now - the intention here was a proof of concept that this setup would work end to end


package.json
```diff
@@ -34,17 +34,21 @@
     "test": "jest",
     "test:coverage": "jest; open ./coverage/lcov-report/index.html",
     "check": "tsc --project ./tsconfig.check.json && pnpm lint",
-    "postinstall": "pnpm i rimraf && pnpm build"
+    "postinstall": "pnpm i rimraf && pnpm build",
+    "generate-types-from-schema": "cd src/export/digitalPlanning/schema/ && json2ts schema.json > types.d.ts"
```
DafyddLlyr, line 38: I did consider making this a script in the subfolder to avoid polluting this here, but it seemed pretty quick / pragmatic.

src/export/digitalPlanning/model.test.ts
```diff
@@ -0,0 +1,103 @@
...
+        passport: mockPassport,
+        templateNames: ["template1", "template2"],
+      });
+
+      const payload = instance.getPayload();
+
+      expect(payload).toEqual(instance.payload);
+    });
+
+    describe("invalid payloads", () => {
```
DafyddLlyr, line 25: I'll admit that this test suite is verging on "let's test an external library" territory but I think it's useful as a sense check that what we're aiming for here will work as expected.

jessicamcinchak (approved):

PR #106, opened 2023-08-22
feat: Get user by email

DafyddLlyr:
Helper method to be used as part of building a JWT for the user.

### DIFF
--- a/src/requests/user.ts
+++ b/src/requests/user.ts
@@ -1,6 +1,8 @@
 import type { GraphQLClient } from "graphql-request";
 import { gql } from "graphql-request";
 
+import { User } from "../types/user";
+
 export class UserClient {
   protected client: GraphQLClient;
 
@@ -18,6 +20,10 @@ export class UserClient {
 
   async _destroy(userId: number): Promise<boolean> {
     return _destroyUser(this.client, userId);
+  }
+
+  async getByEmail(email: string): Promise<User | null> {
+    return getByEmail(this.client, email);
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/types/user.ts
```diff
@@ -0,0 +1,17 @@
...
+}
+
+export interface UserTeams {
+  teamId: number;
+  role: TeamRole;
+}
+
+export type TeamRole = "teamAdmin" | "teamViewer";
+
+export type Role = TeamRole | "public" | "platformAdmin";
```
jessicamcinchak, line 17: I'm curious if it's helpful via types to try to be more clear about what roles we've defined, versus which are Hasura defaults, for example: 
```ts
export type TeamRole = "teamAdmin" | "teamViewer";
export type HasuraRole = "public"; // "public" | "admin" ??
export type Role = TeamRole | HasuraRole | "platformAdmin"; 
```
We might not have enough roles to justify this yet, and it's still a tiny bit confusing here to communicate how `platformAdmin` is a permissions level rather than role, but just thinking ahead !

jessicamcinchak (approved):

src/types/user.ts
DafyddLlyr: This is a great idea - worth being as explicit as we can here in this fuzzy domain between database and logic.

src/requests/user.ts
```diff
@@ -69,3 +75,28 @@ export async function _destroyUser(
...
+  const { users }: { users: User[] | null } = await client.request(
+    gql`
+      query GetUserByEmail($email: String!) {
+        users: users(where: { email: { _eq: $email } }) {
+          id
+          firstName: first_name
+          lastName: last_name
+          email
+          isPlatformAdmin: is_platform_admin
+          teams {
```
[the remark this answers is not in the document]
DafyddLlyr: This is correct - the column is `users.teams`, which is the relationship to the `team_members` table.

To me this naming convention made more sense as "these are the teams the user is a member of" as opposed to "these are the team_member relationships of the user"

image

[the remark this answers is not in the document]
jessicamcinchak: aha I missed that this is a graphql field available via _relationship_, was confused by `team_members.role` column defined directly in table schema - thanks for clarification !

src/types/roles.ts
```diff
@@ -0,0 +1,29 @@
...
+export type HasuraRole = "public";
+
+/**
+ * Roles granted by an authorised user's status, as opposed to their relationship to a team
+ *
+ * platformAdmin: The highest level of permission in PlanX, and can operate across all teams
+ */
+export type UserRole = "platformAdmin";
+
+export type Role = TeamRole | HasuraRole | UserRole;
```
jessicamcinchak, line 29: :100: this is great !

PR #107, opened 2023-08-24
feat: Add/remove team members

DafyddLlyr:
A few basic helper functions to be used via the Swagger docs to make team management a non-dev task in the absence of a UI to allow self-management.

### DIFF
--- a/src/requests/team.ts
+++ b/src/requests/team.ts
@@ -1,6 +1,19 @@
 import type { GraphQLClient } from "graphql-request";
 import { gql } from "graphql-request";
 
+import { TeamRole } from "../types/roles";
+
+interface UpsertMember {
+  userId: number;
+  teamId: number;
+  role: TeamRole;
+}
+
+interface RemoveMember {
+  userId: number;
+  teamId: number;
+}
+
 export class TeamClient {
   protected client: GraphQLClient;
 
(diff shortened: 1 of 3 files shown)


src/requests/team.ts
```diff
@@ -119,3 +144,58 @@ export async function _destroyTeam(
...
+        mutation UpsertTeamMember(
+          $role: user_roles_enum = viewer
+          $team_id: Int
+          $user_id: Int
+        ) {
+          insert_team_members_one(
+            object: { team_id: $team_id, user_id: $user_id, role: $role }
+            on_conflict: {
+              constraint: team_members_user_id_team_id_key
+              update_columns: role
```
DafyddLlyr, line 164: This is super handy - you can't allow more a user to have multiple roles within the same team. This works for the current, very simple, model but may not last forever.

src/requests/team.ts
```diff
@@ -19,6 +32,18 @@ export class TeamClient {
     return createTeam(this.client, args);
   }
 
+  async addMember(args: UpsertMember) {
+    return upsertMember(this.client, args);
+  }
+
+  async changeMemberRole(args: UpsertMember) {
+    return upsertMember(this.client, args);
+  }
```
DafyddLlyr, line 41: Using an upsert here means that "changeMemberRole" is not strictly only able to change an exiting member's role.

Calling it with a new unique combination of teamId/userId will be an INSERT operation but in reality this is going to be used by a handful of experienced users via the API docs for a finite time period, so I don't think this is too much of an issue.

jessicamcinchak (approved):

PR #115, opened 2023-09-01
feat: Improve client auth options

DafyddLlyr:


src/requests/index.ts
```diff
@@ -61,19 +56,6 @@ export class CoreDomainClient {
     this.export = new ExportClient(this.client);
   }
 
-  authorizeSession(sessionDetails: { email: string; sessionId: string }) {
```
DafyddLlyr, line 64: This method of authorising separately to instantiating might be the right pattern, so I'll come back to it if needed. It's possible we could call a function like this in a middleware layer to set the context for the client.

jessicamcinchak (approved):

PR #131, opened 2023-09-18
wip: update digital planning schema types and map to passport

jessicamcinchak:
**Changes:** 
- Copied most recent Open API `schema.json` and generated new types
- Started mapping passport fields to `Payload["data"]` type - this is a bit patchy and tedious as schema is still changing too
- Added mock session data for 4 application types generated by Alastair for testing (only using these for digital planning data right now, but it's a robust set we should use more widely too!)
Screenshot from 2023-09-29 09-01-09

This is enough of a start to satisfy the planx-new admin endpoint work over here https://github.com/theopensystemslab/planx-new/pull/2223

**Next steps:**
- Review & merge this so we have a more stable `generateDigitalPlanningDataPayload` export function on `main` branch
- Repeat :cyclone: (finalise latest round of schema changes, merge those in digital planning repo, copy schema.json back here and keep mapping!)


src/export/digitalPlanning/model.ts
```diff
@@ -31,30 +28,46 @@ export class DigitalPlanning {
...
+          boundary: this.getBoundary(),
+          constraints: {
+            planning: [],
+          },
         },
         application: {
-          name: "Test",
+          type: this.getApplicationType(),
+          fee: this.getApplicationFee(),
+          declaration: {
```
DafyddLlyr, line 57: I appreciate this is the wrong time / place to notice this but presumably we'd need a timestamp here also somewhere?

src/export/digitalPlanning/model.ts
```diff
@@ -64,28 +77,13 @@ export class DigitalPlanning {
         },
         session: {
           source: "PlanX",
-          id: "b3320abe-f5bc-4185-b61f-40e9e65f07ad",
+          id: "b3320abe-f5bc-4185-b61f-40e9e65f07ad", // this.sessionId throws 'must match format UUID' error
```
DafyddLlyr, line 80: Ah this is an interesting one, thanks for the comment

src/export/digitalPlanning/model.ts
```diff
@@ -101,4 +99,160 @@ export class DigitalPlanning {
...
+      };
+    } else {
+      return {
+        role: this.passport.data?.["applicant.siteContact"][0],
+      };
+    }
+  }
+
+  private getApplicationType(): Payload["data"]["application"]["type"] {
+    // @todo figure out how to lookup via exported type or schema.json ??
```
DafyddLlyr, line 173: 👍

DafyddLlyr (approved):
Super clean and easy to read 👍 Looking good.

PR #135, opened 2023-09-21
fix: Explicitly copy `schema.json` after build

DafyddLlyr:
Pointing at commit [8be92e4](https://github.com/theopensystemslab/planx-core/pull/135/commits/8be92e4a75fad346d3f9ffb0731efa089f467f3f) works for me - I can see a the schema file in my `node_modules`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     }
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


jessicamcinchak (approved):
Thanks for this one !!

PR #144, opened 2023-09-29
feat: map passport to Digital Planning Data v0.0.1

jessicamcinchak:


src/requests/session.ts
```diff
@@ -144,6 +145,41 @@ export async function getSessionPassport(
...
+          lowcal_sessions_by_pk(id: $id) {
+            id
+            createdAt: created_at
+            submittedAt: submitted_at
+            flow {
+              id
+              slug
+              publishedFlows: published_flows(
+                order_by: { created_at: desc }
+                limit: 1
```
jessicamcinchak, line 165: note: I'm currently just pulling the latest published flow, _not_ a version that's actually meaningful to this session. 

I realised that the `publishedFlowId` for a session can change over the course of it's active "lifespan", so do we want to:
- pull the `publishedFlowId` based on `submittedAt` (or `createdAt` or last `reconciliation_request` for this session ??)
- only include `flowId`

src/export/digitalPlanning/model.ts
```diff
@@ -254,5 +429,273 @@ export class DigitalPlanning {
...
+      );
+
+      if (this.passport.data?.["proposal.vehicleParking"][0] !== "none") {
+        const vehicleTypes = [
+          "cars",
+          "vans",
+          "motorcycles",
+          "bicycles",
+          "buses",
+        ];
```
DafyddLlyr, line 555: Is there a way we could derive this and `carTypes` from the schema file?

src/export/digitalPlanning/model.ts
```diff
@@ -254,5 +429,273 @@ export class DigitalPlanning {
...
+          );
+        });
+      }
+    }
+
+    return baseProposal;
+  }
+
+  private getFiles(): Payload["files"] {
+    const files: any = [];
```
DafyddLlyr, line 635: This is similar to the above comment - declaring `const files: File[] = [];` here seems worthwhile. We'll need to cast in the function somewhere but it's more likely we'll catch issues by declaring our intentions here rather than just forcing the return type.

DafyddLlyr (changes requested):
A good body of work here! I can really appreciate the back and forth it must have taken to make schema fixes then get the implementation right across here.

Totally support the fix-forward approach so please feel free to address as few or as many comments as you feel is appropriate. The only show stopper is the `Boolean()` coercion which is something worth fixing before we move forward I think.

In terms of testing this is actually great in terms of coverage as running the validation makes sure everything is covered / ran. The only thing I'd probably suggest adding is a literal check on some / all of the validation checks so that we can say - 
 - This session / passport produces a valid `DigitalPlanningApplication` according to the schema
  - The values actually match what we expect. This means we'd catch the `Boolean()` issue described above where the type is correct but value is wrong.

src/requests/session.ts
jessicamcinchak: Let's drop `publlishedFlowId` :+1: 

This is queued up over here: https://github.com/theopensystemslab/digital-planning-data-schemas/pull/57

src/export/digitalPlanning/model.ts
```diff
@@ -254,5 +429,273 @@ export class DigitalPlanning {
...
+        ],
+      );
+    }
+
+    return baseFee;
+  }
+
+  private getApplicationDeclaration(): Payload["data"]["application"]["declaration"] {
+    return {
+      accurate: Boolean(
```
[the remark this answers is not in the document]
jessicamcinchak: Thanks for catching this! Now updated. Because this schema is a bit stricter than OneApp, I've repeated a private function definition that can return a strict `boolean` rather than `boolean | undefined` as OneApp does, while still properly catching/distinguishing between "true" & "false" :+1:

src/export/digitalPlanning/model.ts
jessicamcinchak: I've added a `@todo` code comment to pick up next - I've spent quite a bit of time on this but getting lots of errors still and needing to do a lot of tedious casting, which feels like I'm missing something / a better way.

src/export/digitalPlanning/model.ts
```diff
@@ -218,20 +294,119 @@ export class DigitalPlanning {
...
+        "EPC.number",
+        this.passport.data?.["property.EPC.number"],
+      );
+    }
+
+    return baseProperty;
+  }
+
+  private getPlanningConstraints(): Payload["data"]["property"]["constraints"] {
+    const data: any = [];
```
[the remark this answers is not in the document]
jessicamcinchak: I've added a `@todo` code comment to pick up next - I've spent quite a bit of time on this but getting lots of errors still and needing to do a lot of tedious casting, which feels like I'm missing something / a better way.

src/export/digitalPlanning/model.ts
jessicamcinchak: This is a good question - all the granular variables are a single list under the `VehicleParking` definition so it'll require a bit of parsing to get the right segments, but totally possible. I've added a `@todo` code comment to pick up in the next round of changes.

src/export/digitalPlanning/model.ts
```diff
@@ -40,50 +72,35 @@ export class DigitalPlanning {
...
           source: "PlanX",
-          id: "b3320abe-f5bc-4185-b61f-40e9e65f07ad", // this.sessionId throws 'must match format UUID' error
-          createdAt: "2018-11-13T20:20:39+00:00",
-          submittedAt: "2018-11-13T20:20:39+00:00",
+          id: this.sessionId,
+          createdAt: this.metadata.createdAt,
+          submittedAt: this.metadata.submittedAt,
+        },
+        schema: {
+          url: "https://theopensystemslab.github.io/digital-planning-data-schemas/v0.0.1/schema.json", // @todo populate version based on submittedAt
```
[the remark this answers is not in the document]
jessicamcinchak: Right now, everytime we publish a new schema release, we'll need to update the static schema.json in planx-core too - because we populate some payload values directly from the schema definitions.

So I realised I think this is as simple as a string template that let's us still keep the URL type here:
```https://theopensystemslab.github.io/digital-planning-data-schemas/${jsonSchema["$id"]}/schema.json```

This is a workflow we should definitely keep talking about though! Schema releases and volume of apps are infrequent enough for now that the manual copy step should be manageable.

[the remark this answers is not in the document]
Comment (DafyddLlyr) [reply src/export/digitalPlanning/model.ts]: Ah nice to have such a simple solution! And yep, agree there's fine-tuning on the release process to be done for sure in future.

File: src/export/digitalPlanning/model.ts (hunk 5 above)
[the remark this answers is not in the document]
Comment (DafyddLlyr) [reply src/export/digitalPlanning/model.ts:363]: Yep - sounds like a plan

This PR (https://github.com/theopensystemslab/planx-core/pull/78) also leads to type issues - we should 100% merge this schema first before I take that one further.

File: src/export/digitalPlanning/model.ts [hunk 7]
```diff
@@ -98,7 +96,7 @@ export class DigitalPlanning {
           submittedAt: this.metadata.submittedAt,
         },
         schema: {
-          url: "https://theopensystemslab.github.io/digital-planning-data-schemas/v0.0.1/schema.json", // @todo populate version based on submittedAt
+          url: `https://theopensystemslab.github.io/digital-planning-data-schemas/${jsonSchema["$id"]}/schema.json`,
```
DafyddLlyr, line 99: 👏

src/export/digitalPlanning/model.ts
```diff
@@ -206,6 +215,8 @@ export class DigitalPlanning {
       siteContact: this.getSiteContact(),
     };
 
+    // Prior Approval application types don't collect additional ownership info
```
DafyddLlyr, line 218: 👍

DafyddLlyr (approved):
Happy with all changes and `@todo`s 🎉

PR #162, opened 2023-10-16
refactor: Make redaction explicit in `ExportClient`

DafyddLlyr:
## What does this PR do?
This PR makes redaction of export payloads explicit. Currently we always generate both a redacted and non-redacted payload per request which needlessly calls an expensive operation twice. The aim here is to make PlanX submissions run a bit more smoothly and quickly.

Implemented in https://github.com/theopensystemslab/planx-new/pull/2313

## Also...
- Standardise types in the `ExportClient`
- Moves `generateOneAppXML()` from requests to export


jessicamcinchak (approved):
A very good tidy up! :crossed_fingers: performance gains extend beyond local testing next

PR #176, opened 2023-10-26
feat: update digital planning data schema mappings

jessicamcinchak:
- adds schema v0.1.2 & generates new types
- improves passport mapping throughout (especially more conditional checks to omit optional properties rather than include as `{}`)

### DIFF
--- a/src/export/digitalPlanning/model.ts
+++ b/src/export/digitalPlanning/model.ts
@@ -23,6 +23,7 @@ import {
   DigitalPlanningApplication as Payload,
   File,
   FileType,
+  GeoJSON,
   LondonProperty,
   PlanningConstraint,
   ProjectType,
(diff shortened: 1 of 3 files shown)


src/export/digitalPlanning/model.ts
```diff
@@ -323,7 +313,9 @@ export class DigitalPlanning {
 
   private getBoundary(): Payload["data"]["property"]["boundary"] {
     return {
-      site: this.passport.data?.["property.boundary.site"],
+      site: this.passport.data?.[
+        "property.boundary.site"
+      ] as unknown as GeoJSON,
```
jessicamcinchak, line 318: `as unknown as GeoJSON` is an especially odd casting but it seems to be what TS wanted! Am I missing something obvious?

DafyddLlyr: Yep I also had to do this elsewhere - it's the best we can do for now and is correct! 👌

DafyddLlyr: (+ it's whatever value we put in here is going to get parsed and validated shortly anyway)

DafyddLlyr (approved):

PR #211, opened 2023-11-29
fix: update passport <> ODP Schema mappings

jessicamcinchak:
Was checking recent application payloads today (eg [admin endpoint for BOPS](https://api.editor.planx.uk/admin/session/6508e9bd-a50e-4621-b6d9-f9cd3e23a214/bops) compared to [admin endpoint for schema](https://api.editor.planx.uk/admin/session/6508e9bd-a50e-4621-b6d9-f9cd3e23a214/digital-planning-application) for the most recent Southwark LDC) and I noticed some unexpected 0s and undefineds in the new schema mapping. 

Confusingly, these still validate successfully but are false data - it's easy to overlook!

It seems most of these were introduced through the use of `generic<>` which I've removed here for now, in favor of the clunkier syntax `as <type>`. Very open to other suggestions.

Examples in the ODP schema repo are okay as I've been updating them section-by-section for a bit now (eg just log out new `planning.designations` shape from model.test.ts here and copy over to examples in other repo rather than full payload), but this also meant that this mapping bug has gone unnoticed for awhile.

Please take a look at the admin endpoints and shout if you spot any other inconsistencies not addressed here! We'll start sending to BOPS soon and I'm keen to make sure data mappings and validation are tight on our side.

### DIFF
--- a/src/export/digitalPlanning/model.ts
+++ b/src/export/digitalPlanning/model.ts
@@ -360,7 +360,11 @@ export class DigitalPlanning {
       }),
     };
 
-    if (this.passport.data?._address?.["property.region"]?.[0] === "London") {
+    // Prior Approvals will use London Data Hub in future, but don't yet https://editor.planx.uk/opensystemslab/prior-approval-more-information
+    if (
+      this.passport.data?.["property.region"]?.[0] === "London" &&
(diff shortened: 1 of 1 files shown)


DafyddLlyr (approved):
Thanks for catching this - I'll aim to get back to the passport class and implement a proper `get()` method with an abundance of unit tests!

PR #241, opened 2023-12-22
chore: Bump dependencies

DafyddLlyr:


package.json
```diff
@@ -44,9 +44,9 @@
     "copy-json-schema": "copyfiles -f ./src/export/digitalPlanning/schema/schema.json ./dist/export/digitalPlanning/schema"
   },
   "dependencies": {
-    "@emotion/react": "^11.11.1",
+    "@emotion/react": "^11.11.3",
     "@emotion/styled": "^11.11.0",
-    "@mui/material": "^5.14.20",
+    "@mui/material": "^5.15.2",
```
jessicamcinchak, line 49: nit: if I remember right, we want material ui versions to stay in sync between these two repos else we can get opaque CI bugs - can you please bump planx-new mui version too whenever you open the PR to bump planx-core? https://github.com/theopensystemslab/planx-new/blob/main/editor.planx.uk/package.json#L13

jessicamcinchak (approved):

package.json
DafyddLlyr: Good catch - will do..!

PR #242, opened 2023-12-22
chore: Update `team.create()` to account for new `team_themes` table relationship

DafyddLlyr:
See https://github.com/theopensystemslab/planx-new/pull/2602

### DIFF
--- a/src/requests/team.ts
+++ b/src/requests/team.ts
@@ -113,7 +113,7 @@ export async function createTeam(
     slug,
     theme: {
       logo,
-      primary: primaryColor,
+      primary_colour: primaryColor,
     },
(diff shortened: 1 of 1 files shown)


src/requests/team.ts
```diff
@@ -113,7 +113,7 @@ export async function createTeam(
     slug,
     theme: {
       logo,
-      primary: primaryColor,
+      primary_colour: primaryColor,
```
jessicamcinchak, line 116: nit: `primary_colour` & `primaryColor` feel easily to confuse here - can we consistently use one or the other?

jessicamcinchak (approved):

PR #269, opened 2024-01-23
chore: account for title boundaries in document templates

jessicamcinchak:
Since introducing title boundaries, we should ensure that any document templates that display the boundary: 1) have proper attribution and 2) explain the action the user took to planning officers.

`Overview.html` & `Map.html` will now always show a "Source" underneath the map image, and will conditionally render the second attribution title boundary on the map if the user accepted or amended it as their location plan. 

For example:
Screenshot from 2024-01-24 11-22-51

Screenshot from 2024-01-24 10-27-17

### DIFF
--- a/src/export/digitalPlanning/model.ts
+++ b/src/export/digitalPlanning/model.ts
@@ -1,5 +1,6 @@
 import Ajv from "ajv";
 import addFormats from "ajv-formats";
+import { Feature } from "geojson";
 import set from "lodash.set";
 
 import { Passport } from "../../models";
@@ -341,10 +342,15 @@ export class DigitalPlanning {
   }
 
   private getProposedBoundary(): Payload["data"]["proposal"]["boundary"] {
+    const annotatedBoundary = this.passport.data?.[
+      "property.boundary.site"
+    ] as unknown as Feature;
+    if (annotatedBoundary && annotatedBoundary.properties)
+      annotatedBoundary["properties"]["planx_user_action"] =
+        this.passport.data?.["drawBoundary.action"];
+
     return {
-      site: this.passport.data?.[
-        "property.boundary.site"
-      ] as unknown as GeoJSON,
+      site: annotatedBoundary as GeoJSON,
       area: {
         hectares:
           this.passport.data?.["proposal.siteArea.hectares"] ||
--- a/src/export/digitalPlanning/model.ts
+++ b/src/export/digitalPlanning/model.ts
@@ -342,9 +342,7 @@ export class DigitalPlanning {
 
   private getProposedBoundary(): Payload["data"]["proposal"]["boundary"] {
     return {
-      site: this.passport.data?.[
-        "property.boundary.site"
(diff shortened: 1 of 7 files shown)


src/templates/html/map/MapHTML.tsx
```diff
@@ -18,7 +19,14 @@ export function MapHTML(props: {
...
         <h1>Boundary</h1>
-        <Map boundary={props.geojson} clipGeojsonData={props.boundingBox} />
+        <Map
+          boundary={props.geojson}
+          clipGeojsonData={props.boundingBox}
+          userAction={props.userAction}
+        />
+        {props.userAction && (
+          <p>This PlanX user {props.userAction.toLowerCase()}.</p>
+        )}
```
jessicamcinchak, line 29: Conditional examples based on user action: 
Screenshot from 2024-01-23 16-12-36

(Second attribution line isn't displaying here like it should, need to debug this one - might be different style css?)
Screenshot from 2024-01-23 16-51-44

src/export/digitalPlanning/model.ts
```diff
@@ -341,10 +342,15 @@ export class DigitalPlanning {
   }
 
   private getProposedBoundary(): Payload["data"]["proposal"]["boundary"] {
+    const annotatedBoundary = this.passport.data?.[
+      "property.boundary.site"
+    ] as unknown as Feature;
+    if (annotatedBoundary && annotatedBoundary.properties)
+      annotatedBoundary["properties"]["planx_user_action"] =
+        this.passport.data?.["drawBoundary.action"];
+
```
jessicamcinchak, line 351: small schema-related change snuck in here too: we had already mapped the title boundary passport variable to populate `data.property.boundary` & the location plan to populate `data.proposal.boundary` - but this takes it one step further and adds a custom GeoJSON `properties` key onto the `data.proposal.boundary` to explain which action the Planx user took. In cases of amend or accept, `properties` will additionally include the entity source info from Digital Land.

we'll be overdue soon to re-run some example payloads & populate fresh/more realistic example payloads for new features like this.

src/templates/html/application/ApplicationHTML.tsx
```diff
@@ -252,7 +255,7 @@ export function ApplicationHTML(props: {
       <head>
         <meta charSet="utf-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1" />
-        <script src="https://cdn.jsdelivr.net/npm/@opensystemslab/map@0.7.5"></script>
+        <script src="https://cdn.jsdelivr.net/npm/@opensystemslab/map@0.7.9"></script>
```
DafyddLlyr, line 258: I'm certainly guilty of missing this version bump! Good catch.

DafyddLlyr (approved):

src/templates/html/application/ApplicationHTML.tsx
jessicamcinchak: Same - surprisingly wasn't that far back!

PR #271, opened 2024-01-24
fix: map props need to include style and buffers should be numbers

jessicamcinchak:
It failed on the `style` tag :weary: 

Also cleaning up all buffer types which we use a bit inconsistenly between numbers/strings - Lit evaluates same / smartly but TS is picky !

### DIFF
--- a/src/templates/html/map/Map.tsx
+++ b/src/templates/html/map/Map.tsx
@@ -29,7 +29,7 @@ declare global {
       geojsonData?: string;
       geojsonColor?: string;
       geojsonFill?: boolean;
-      geojsonBuffer?: string;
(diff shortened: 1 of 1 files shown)


src/templates/html/map/Map.tsx
```diff
@@ -1,4 +1,5 @@
 import * as React from "react";
+import { any, string } from "zod";
```
DafyddLlyr, line 2: Is this correct or an accidental import?

jessicamcinchak: accidental - will remove - good catch !

DafyddLlyr (approved):

PR #272, opened 2024-01-25
fix: ensure boundary HTML doesn't display print dialog

jessicamcinchak:
The MapHTML was missing these OL CDNs, while the ApplicationHTML had them. Consistently using among both generated documents means the print dialog behaves consistently ! 

Before:
Screenshot from 2024-01-24 19-37-00

After:
Screenshot from 2024-01-25 11-06-07


DafyddLlyr (approved):
Nice and simple! Thanks for picking this one up.

PR #273, opened 2024-01-25
feat: bump `@opensystemslab/map` to v0.8.0

jessicamcinchak:
Will bump planx-core & map in one go over in planx-new once this merges because of type dependency, that will also pick up the boundary.html fix from earlier !


DafyddLlyr (approved):

DafyddLlyr (commented):
@jessicamcinchak I think we need a version bump in `Application.HTML.tsx` also right?

PR #281, opened 2024-01-29
feat: `team.getIntegrations()` function

DafyddLlyr:


src/requests/team.ts
```diff
@@ -271,34 +273,55 @@ async function getBySlug(client: GraphQLClient, slug: string) {
...
+interface Integrations {
+  bopsSubmissionURL?: string;
+  bopsToken?: string;
+}
+
+/**
+ * Return integration details for a team
+ *
+ * XXX: Why do we need env? Why are there columns for staging and production secrets?
+ * Please see https://github.com/theopensystemslab/planx-new/pull/2499
```
DafyddLlyr, line 294: It would be good to talk about this at the next dev call - I've added an item to the agenda.

src/utils/encryption.ts
```diff
@@ -0,0 +1,44 @@
+import * as crypto from "crypto";
+
+/**
+ * Encrypts a secret using AES-256-CBC encryption
+ *
+ * @param secret - The secret to be encrypted
+ * @param key - The encryption key - a 32-byte string
+ * @returns The encrypted secret along with the initialization vector (IV), separated by a colon
+ */
+export function encrypt(secret: string, key: string): string {
```
DafyddLlyr, line 10: Inspiration / instruction for these functions came from here - https://www.geeksforgeeks.org/node-js-crypto-createcipheriv-method/

I had originally planned to use a library for this, but it looks like the move is towards doing this natively in Node - `[crypto-js](https://www.npmjs.com/package/crypto-js)` is no longer maintained and advises using the Node crypto module directly.

src/requests/team.ts
```diff
@@ -271,34 +273,55 @@ async function getBySlug(client: GraphQLClient, slug: string) {
...
 }
 
-async function getBopsSubmissionURL(
-  client: GraphQLClient,
-  slug: string,
-  env: PlanXEnv,
-) {
+interface Integrations {
+  bopsSubmissionURL?: string;
+  bopsToken?: string;
```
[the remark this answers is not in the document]
DafyddLlyr: Yep - good call - explicit is better!

The naming convention I'm aiming for with columns is `*_secret` and then in this `DecryptedIntegrations` interface we can use the "real" name (this might be GovPay API key, Uniform client, Bops Token etc)

PR #283, opened 2024-01-31
feat: Create empty `team_theme` and `team_integrations` records on team creation

DafyddLlyr:
## What does this PR do?
 - Create an empty `team_integrations` record on team creation
 - Create an empty `team_theme` record on team creation
 - Update arguments for `team.create()` to better reflect what is generally available at time of creation - make submission email optional. Boundary and reference code are also optional as not all teams will be local authorities (e.g. Templates, OSL, ODP, WikiHouse)

Implemented in https://github.com/theopensystemslab/planx-team-onboarding/pull/16 and https://github.com/theopensystemslab/planx-new/pull/2720

### DIFF
--- a/src/requests/team.ts
+++ b/src/requests/team.ts
@@ -1,3 +1,4 @@
+import { GeoJsonObject } from "geojson";
 import type { GraphQLClient } from "graphql-request";
 import { gql } from "graphql-request";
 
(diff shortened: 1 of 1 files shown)


src/requests/team.ts
```diff
@@ -18,21 +19,23 @@ interface RemoveMember {
 
 type PlanXEnv = "pizza" | "staging" | "production";
 
+interface CreateTeam {
```
DafyddLlyr, line 22: We could also make `theme` an optional argument here in future if we asked this in the future when teams can be created within PlanX

jessicamcinchak (approved):

PR #290, opened 2024-02-05
feat: Read file details from `Passport()` class

DafyddLlyr:
## What does this PR do?
- Uses a single method to read files from the passport
- Quick refactor before working on the wider file-type changes to add requested files to the ODP schema payload

### DIFF
--- a/src/export/digitalPlanning/model.ts
+++ b/src/export/digitalPlanning/model.ts
@@ -716,7 +716,7 @@ export class DigitalPlanning {
   private getFiles(): Payload["files"] {
     const files: File[] = [];
 
-    this.passport.fileDetails().forEach(({ url, key }) => {
+    this.passport.files.forEach(({ url, key }) => {
       try {
         // push a new label to an existing file
         if (files.filter((file) => file.name === url).length > 0) {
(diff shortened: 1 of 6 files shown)


src/models/passport/passport.ts
```diff
@@ -1,32 +1,100 @@
 import { get as getByKeyPath, has } from "lodash";
 
+// import { ServiceFiles } from "../../export/digitalPlanning/schema/types";
```
DafyddLlyr, line 3: These are intentionally commented out to allow this PR to proceed without a schema update - they'll be commented back in shortly.

src/export/digitalPlanning/model.ts
```diff
@@ -716,42 +716,38 @@ export class DigitalPlanning {
...
-                  key,
-                ),
-              });
-            }
-          } catch (err) {
-            throw new Error(`Error formatting files: ${err}`);
-          }
-        });
-      });
+    this.passport.files.forEach(({ url, key }) => {
```
jessicamcinchak, line 719: one small question: can we confirm that we'll still pick up location plan files if we read from `files` rather than `data`? I don't see a test case covering that? It's a bit of an edge case as it's the only way to upload a file outside of an upload component.

jessicamcinchak (approved):
Looks good, one small non-blocking question. 

Thanks for tidy up!

src/export/digitalPlanning/model.ts
DafyddLlyr: Very good shout - will add a test case 👍

PR #302, opened 2024-02-13
fix: unskip schema tests

jessicamcinchak:
These were temporarily skipped while sequencing merges about document-type changes and we forgot to re-enable !


DafyddLlyr (approved):

DafyddLlyr:
Good catch!

PR #309, opened 2024-02-22
feat: include the ODP Schema as "application.json" in files sent to Uniform

jessicamcinchak:
Related to https://github.com/theopensystemslab/planx-new/pull/2804

Rather than making this an optional argument on the `buildSubmissionExportZip()` method like `includeDigitalPlanningSchemaJSON` - we can just include by default for supported application types (Uniform = always LDC) which will give councils more flexibility (eg retrieve the JSON from DMS rather than email inbox; not have to configure a service to send to all three possible destinations) and communicates to Idox that we're ready/able to be sending this new data format regardless of Nexus.

Once this is approved and merged to `main` along with 2804, I'll submit a test LDC on staging and get a council PO to confirm it can be received correctly via DMS and doesn't break before going to production.

### DIFF
--- a/src/export/oneApp/OneApp.test.ts
+++ b/src/export/oneApp/OneApp.test.ts
@@ -601,6 +601,10 @@ describe("File handling", () => {
         "common:Reference": "Other",
       },
       {
+        "common:FileName": "application.json",
+        "common:Reference": "Other",
+      },
+      {
         "common:Identifier": "N10049",
         "common:FileName": "proposal.xml",
         "common:Reference": "Schema XML File",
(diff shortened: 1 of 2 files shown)


jessicamcinchak:
[Re-evaluate how "Send to email" works (in light of PP services primarily using this until other integrations are ready/adopted)](https://trello.com/c/3HW39zX6/2757-re-evaluate-how-send-to-email-works-in-light-of-pp-services-primarily-using-this-until-other-integrations-are-ready-adopted)

Mike-Heneghan (approved):
Great plan 👍

PR #310, opened 2024-02-23
Revert "feat: include the ODP Schema as "application.json" in files sent to Uniform"

jessicamcinchak:
Reverts theopensystemslab/planx-core#309

Doesn't look hopeful that DMS will accept a JSON, and found out most folks' staging envs were either turned off or inaccessible and this isn't a necessary change, so probably not worth the test trouble.


Mike-Heneghan (approved):
👍

PR #334, opened 2024-03-19
chore: Add `govPayMetadata` property to `PaymentRequest` type

DafyddLlyr:


DafyddLlyr:
@jessicamcinchak Made some substantial changes here since your review sorry - could you take another look whenever you get a chance?

src/requests/payment-request.ts
```diff
@@ -35,7 +35,13 @@ export class PaymentRequestClient {
 
 type PayNode = {
   type: ComponentType.Pay;
-  data: { fn: string | undefined };
+  data: {
+    fn: string | undefined;
```
jessicamcinchak, line 39: In the Editor, I believe `fn` is required for Pay - can it be here too?

jessicamcinchak (approved):
Changes look good, extra docs appreciated - one small question, not a blocker!

src/requests/payment-request.ts
DafyddLlyr: Good catch! Resolved by [755eef](https://github.com/theopensystemslab/planx-core/pull/334/commits/3755eef3715b99a71f74c7efc3c8a837ae7cc9ca)

PR #339, opened 2024-03-27
chore: update `parking` mappings in ODP Schema

jessicamcinchak:


src/export/digitalPlanning/model.ts
```diff
@@ -411,12 +423,31 @@ export class DigitalPlanning {
...
 
+  private getExistingParking(): LondonProperty["parking"] {
+    const parking = {};
+    PARKING_TYPES.forEach((type) => {
+      if (type === "offStreet.residential") {
+        parking["offStreet"] = {
+          residential: {
+            count: this.passport.data?.[`property.parking.${type}`] || 0,
+          },
+        };
```
jessicamcinchak, line 441: What's the clever syntax I'm missing here to be able to pass a dot-separated string to create a new nested object property?

src/export/digitalPlanning/model.ts
```diff
@@ -591,134 +622,51 @@ export class DigitalPlanning {
...
+          baseProposal,
+          `parking.${type}.count`,
+          this.passport.data?.[`proposal.parking.${type}`] || 0,
+        );
+        set(
+          baseProposal,
+          `parking.${type}.difference`,
+          this.passport.data?.[`proposal.parking.${type}.difference`] || 0,
+        );
+      });
```
jessicamcinchak, line 669: We have a bit on an interesting predicament here - 
- We have a new schema structure that counts certain types of parking that we want to fulfill on payload generation
- I've updated the passport mappings to reflect the _new_ content that has been written to capture this
- Even once the new content is published in the London Data Hub, there's going to be a time lag while parent services get published and existing sessions could still ask/capture the _old_ passport variables 
- At first I thought we'd perhaps be able to have multiple levels of fallbacks here (eg `"new.passport.var" || "old.passport.var" || 0`), but the structures are actually _counting different things_ and I'm not sure we can make that comparison so simply??

So, are we okay understanding that any services slow to publish new LDH changes won't have structured parking data, but those questions and answers would still come through in `responses` ? BOPS was quite clear they aren't doing anything with this yet, so hopefully no major issues?

src/export/digitalPlanning/model.ts
DafyddLlyr: I think what you're looking for is `set()` from Lodash ([docs](https://lodash.com/docs/4.17.15#set))

Example in out code here - https://github.com/theopensystemslab/planx-core/blob/main/src/export/digitalPlanning/model.ts#L600-L604

src/export/digitalPlanning/model.ts
DafyddLlyr: Yeah this is an interesting one and one that will appear a few times no doubt.

If we're counting different things, the fallback isn't a viable option I agree. I think as the data isn't currently being consumed, the approach outlined here is the right way forward 👍 

Hypothetically, if BOPS were consuming this, it gets more interesting!

I think we'd have to make it initially an optional field, wait until services updated, and then update schema to make it mandatory once that was done? Or ensure all services are updated first (hard to coordinate!) and then roll out changes to schema?

I think in an ideal world we'd want it always to be schema led. If BOPS need something, we add to schema, and then PlanX services have to "catch up" and fail validation/publishing (somehow) until they're compliant. An interesting one to mull over.

DafyddLlyr: Slightly tangential, but I was using this query the other day to chec kout Pay nodes which didn't have an update value for `govPayMetadata`.

Helpful pattern for checking if values are set, could be applied to `fn` fields in future?

```sql
SELECT 
    obj.key AS pay_key, 
    obj.value AS pay_node,
    flows.id as flow_id
FROM 
    flows,
    jsonb_each(flows.data) AS obj
WHERE 
    jsonb_typeof(obj.value) = 'object'
    AND (obj.value->>'type')::int = 400
    AND (obj.value->'data'->>'govPayMetadata') IS NULL;
```

DafyddLlyr (approved):

src/export/digitalPlanning/model.ts
jessicamcinchak: Yep updated throughout now - was just being avoidant to re-writing this section of mappings using `set()` in case there was another route - but this is definitely cleanest :+1:

DafyddLlyr (approved):

PR #348, opened 2024-04-12
feat: `formatGovPayMetadata()` helper

DafyddLlyr:
## What does this PR do?
 - Adds a `formatGovPayMetadata()` helper function with two responsibilities - 
   - Format "Editor" data shape used in Edtior modal and saved in `flows` and `payment_request` tables (`Record<string, string | boolean | number>[]`) to data shape required by GovPay API (`Record<string, string>`)
   - Validate that any dynamic variables (taken from Passport) ahere to the validation rules[ outlined by GovPay here](https://docs.payments.service.gov.uk/reporting/#add-more-information-to-a-payment-39-custom-metadata-39-or-39-reporting-columns-39)
 - Moves shared types to `planx-core`

## Motivation
Currently, we have a simpler helper function (`formatMetadata()`) duplicated in Editor and API projects - it's a one-liner that just transforms from shape to shape. 

As we now need the additional logic to both read from the passport, and validate, it makes sense to write and test this once.

Implemented in https://github.com/theopensystemslab/planx-new/pull/3015

### DIFF
--- a/src/utils/govPayMetadata.test.ts
+++ b/src/utils/govPayMetadata.test.ts
@@ -1,0 +1,119 @@
+import type { GovPayMetadata, Passport } from "../types";
+import { formatGovPayMetadata } from "./govPayMetadata";
+
+const mockMetadata: GovPayMetadata[] = [
+  { key: "firstKey", value: "firstValue" },
+  { key: "secondKey", value: "secondValue" },
+  { key: "key_string", value: "@string" },
+  { key: "key_stringGreaterThan100Chars", value: "@stringGreaterThan100Chars" },
+  { key: "key_stringArray", value: "@stringArray" },
+  {
+    key: "key_stringArrayGreaterThan100Chars",
+    value: "@stringArrayGreaterThan100Chars",
+  },
+  { key: "key_number", value: "@number" },
+  { key: "key_object", value: "@object" },
+  { key: "key_boolean", value: "@boolean" },
+  {
+    key: "key_someValueNotSetInPassport",
+    value: "@someValueNotSetInPassportolean",
+  },
+];
+
+const mockPassport: Passport = {
+  data: {
+    string: "agent",
+    stringGreaterThan100Chars:
+      "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Phasellus volutpat ex a tempor facilisis. Nullam eget mi velit.",
+    stringArray: ["new.building", "new.solarPanel"],
+    stringArrayGreaterThan100Chars: [
+      "Lorem ipsum dolor sit amet",
+      "consectetur adipiscing eli",
+      "Phasellus volutpat ex a tempor facilisis",
+      "Nullam eget mi velit",
+    ],
+    number: 123,
+    object: { someGeoJSON: { abc: 123 } },
+    boolean: true,
+  },
+};
+
+describe("transforming the payload to the correct format for GovPay", () => {
+  it("handles variables set in the Editor", () => {
+    const result = formatGovPayMetadata(mockMetadata, mockPassport);
+    expect(result).toMatchObject({
+      firstKey: "firstValue",
+      secondKey: "secondValue",
+    });
+  });
+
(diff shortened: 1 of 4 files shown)


src/utils/govPayMetadata.ts
```diff
@@ -0,0 +1,91 @@
...
+        : `Error: Invalid metadata value for "${key}" set in PlanX`;
+      break;
+    default:
+      validated = `Error: Invalid metadata value for "${key}" set in PlanX`;
+      break;
+  }
+
+  // Truncate to max 100 characters
+  // TODO: Should we truncate or throw error here?
+  if (typeof validated === "string") validated = validated.substring(0, 100);
```
DafyddLlyr, line 70: I'm actually more inclined to pass back an error here if the value exceeds the allowed length.

I suspect that an error is a probably more meaningful than a truncated value - maybe one for POs to answer?

src/utils/govPayMetadata.test.ts
```diff
@@ -0,0 +1,119 @@
...
+    );
+  });
+
+  it("returns an error for objects", () => {
+    const [_, result] = get("object");
+    // Error passed to GovPay API as string, payment allows to proceed
+    expect(result).toMatch(
+      /Error: Invalid metadata value for "key_object" set in PlanX/,
+    );
+  });
```
jessicamcinchak, line 106: I'll be curious how quickly this one comes up as a limitation - specifically anticipating `_address.uprn` - but very happy for it to throw an error initially !

src/utils/govPayMetadata.ts
jessicamcinchak: I'd actually vote truncate I think ! Just ideally in an obvious way (eg using `...`) 

If throwing an error, do you actually anticipate POs are ready to self-troubleshoot & fix or would this just immediately bounce up in help-issues? I'm (perhaps too negatively!) imagining the latter.

jessicamcinchak (approved):
Looks good, helpful tests !

src/utils/govPayMetadata.test.ts
DafyddLlyr: Yeah this is fair, it may well be something users ask for / expect. It's really just hard coded variables like `_address` that are objects really I think.

This could actually be allowed with recursion here - I'll take a look, it might actually not be a big change to allow this to work.

src/utils/govPayMetadata.ts
DafyddLlyr: Using an ellipsis is actually a very good idea - I'll make this change.

My reasoning for starting the string with `"Error:` would be that it should be immediately obvious to the person who has GovPay access that something is wrong here, whilst a truncated string might not obviously be wrong (maybe unlikely, but an array of values that ended up being cut to 100 characters without any obvious errors could not get spotted).

I think we go with an ellipsis for now, and wait for feedback (and also raise at PO testing). The use cases for this feature currently are to get a number only so this may not really be a relevant issue.

DafyddLlyr: Done! See [1923f12](https://github.com/theopensystemslab/planx-core/pull/348/commits/1923f12660404b90b6a56cca35127f71bb18cfaa)

src/utils/govPayMetadata.test.ts
DafyddLlyr: Yeah also curious if / when this comes up!

The main issue we face here is the issue of key paths when accessing passport variables within objects, when the passport is a combination a of nested and flattened structure.

For example, if we wanted the value for `_address.uprn` we can't say `passport.data["_address.uprn"]`, we'd need to do `passport.data["_address"]["uprn"]`. Likewise, if we use a key path (like we do with the getters in the `Passport` class) we'd need to know where to split `_address.uprn`, but not `proposal.projectTypes`.

It's certainly a solvable problem, (check for existence of value, iterate over key splits) but not a trivial one. We'll need to tackle this when looking at embedding passport variables more widely within the Editor.

PR #352, opened 2024-04-15
fix: Ensure default value for metadata are populated to DB

DafyddLlyr:
Implemented in https://github.com/theopensystemslab/planx-new/pull/3015

This PR ensures that default values are populated into the database when a payment request is created, as required. Previously this was being done on the API and defaults were being sent to GovPay only.

### DIFF
--- a/src/requests/payment-request.ts
+++ b/src/requests/payment-request.ts
@@ -37,7 +37,7 @@ type PayNode = {
   type: ComponentType.Pay;
   data: {
     fn: string;
-    govPayMetadata: {
+    govPayMetadata?: {
       key: string;
       value: string | boolean;
     }[];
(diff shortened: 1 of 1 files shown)


src/requests/payment-request.ts
```diff
@@ -85,7 +85,12 @@ export async function createPaymentRequest(
   // payment requests can only be created for flows with a pay component
   // this throws if the pay component cannot be found
   const payNode = await getPayNode(client, session);
-  const govPayMetadata = payNode.data.govPayMetadata;
+  const govPayMetadata = payNode.data?.govPayMetadata || [
+    { key: "source", value: "PlanX" },
+    { key: "inviteToPay", value: true },
```
jessicamcinchak, line 90: nit: should the default value for `inviteToPay` be `false` since we know it's seldom used currently ?

jessicamcinchak (approved):

src/requests/payment-request.ts
DafyddLlyr: Good question! In this instance, `true` is correct as this will be the value used for a `payment_request` record as they're used for invite to pay journeys only.

PR #388, opened 2024-05-20
feat: Setup `flow.setStatus()` function

DafyddLlyr:
Please see https://github.com/theopensystemslab/planx-new/pull/3170 for context.

This PR adds a helper method to toggle the status of the flow. This is used in E2E tests, and will be called by Editors in the frontend.

### DIFF
--- a/src/requests/flow.ts
+++ b/src/requests/flow.ts
@@ -2,7 +2,7 @@ import type { GraphQLClient } from "graphql-request";
 import { gql } from "graphql-request";
 import { capitalize } from "lodash";
 
-import type { FlowGraph } from "../types";
+import type { FlowGraph, FlowStatus } from "../types";
 
 export class FlowClient {
   protected client: GraphQLClient;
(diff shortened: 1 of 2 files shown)


src/requests/flow.ts
```diff
@@ -264,3 +268,44 @@ export async function _destroyPublishedFlow(
...
+async function setStatus(
+  client: GraphQLClient,
+  args: { flow: { id: string }; status: FlowStatus },
+) {
+  try {
+    const { flow } = await client.request<SetFlowStatus>(
+      gql`
+        mutation SetFlowStatus(
+          $flowId: uuid!
+          $status: flow_status_enum_enum!
```
jessicamcinchak, line 288: `_enum_enum` looks like a possible typo?

jessicamcinchak (approved):

src/requests/flow.ts
DafyddLlyr: This is actually correct! The table has the suffix `_enum` and then the type Hasura generates adds it on again annoyingly. This happens a few other places, e.g. https://github.com/theopensystemslab/planx-new/blob/main/api.planx.uk/modules/send/utils/helpers.ts#L86

I think the solution is not to name enum tables `*_enum`?

PR #390, opened 2024-05-21
fix: account for manually answered planning constraints in ODP Schema generation

jessicamcinchak:
See thread here (this has only occured on test env, not production): https://opendigitalplanning.slack.com/archives/C0241GWFG4B/p1715963778547149

Previously our passport mappings only checked for `_constraints` key which is populated based on Planning Data fetch.

Now this adds an extra fallback check to map from `property.constraints.planning` in cases of manual user answers (eg see this session's passport: https://api.editor.planx.dev/admin/session/74068ec3-2e9c-408a-ac6f-7832d1eb011f/summary)

### DIFF
--- a/src/export/digitalPlanning/model.ts
+++ b/src/export/digitalPlanning/model.ts
@@ -496,58 +496,88 @@ export class DigitalPlanning {
 
   private getPlanningConstraints(): Payload["data"]["property"]["planning"] {
     const teamSlug: string = this.metadata.flow.team.slug;
-    const constraints = this.passport.data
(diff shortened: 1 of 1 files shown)


jessicamcinchak:
Ongoing issues with Github Actions impacting CI here https://www.githubstatus.com/

DafyddLlyr (approved):
Looks good - useful to cover this edge case.

Might be nice to have some test coverage here, but no need to hold up for this (maybe as simple as dropping `_contrainsts` from an existing mock?)

PR #410, opened 2024-06-10
feat: Adding Name Column Changes

RODO94:
# Added FlowName to variables for a flow

To match changes I have made in ``planx-new`` I have added the ``name`` column from the ``flows`` table to the shapes and functions defined here

This has been in order to fix E2E testing failures on ``planx-new`` to do with *rory/add-name-editor-2* changes.


jessicamcinchak (approved):
Looks good to me !

PR #412, opened 2024-06-13
feat: Adding "Name" to the "Session" type for Email Templates 

RODO94:
# What does this PR do?

``InviteToPayData`` & ``InviteToPayMocks`` from **planx-new** are using the ``Session`` type when building the variables for the email templates. In order to use the new ``name`` column as the ``sessionName`` I need to update this ``Session`` type.


jessicamcinchak (approved):
Looks good !

PR #424, opened 2024-06-25
feat: Adding ``team_settings`` to requests and types

RODO94:
## What does this PR do?

Following work done on ``planx-new`` with the addition of a new ``team_settings`` table in the database and form in the Editor. Changes need to be made to planx-core types and requests so we can get the new team general settings from the database when retrieving teams with their ``$teamSlug``


RODO94:
@DafyddLlyr  I have trialled this in my ``planx-new`` PR https://github.com/theopensystemslab/planx-new/pull/3319

src/types/team.ts
```diff
@@ -32,6 +33,17 @@ export interface TeamSettings {
   boundary?: string;
 }
 
+export interface GeneralTeamSettings {
```
DafyddLlyr, line 36: I'd go for the simpler / clearer `TeamSettings` throughout. 

This will simplify things for us as developer because -
 - it matches the database / data model
 - it means we can update the existing `TeamSettings` in `planx-new` instead of maintaining both a `TeamSettings` and `GeneralTeamSettings`
 - "General" is not semantically meaningful enough

I agree that from a user's POV this may not align (e.g. Team Settings > Settings is a bit unclear), but I think we can make this distinction of "General settings" purely at the presentation layer as opposed to the logic layer.

src/requests/team.ts
```diff
@@ -263,6 +270,16 @@ async function getBySlug(client: GraphQLClient, slug: string) {
             logo
             favicon
           }
+          team_settings {
```
DafyddLlyr, line 273: Have a look at all places where `getBySlug()` is used - it may not be the appropriate place to get the settings from. 

If `getBySlug()` is used in a public context (I think it is), the request may fail as the public role cannot (and should not) access team settings.

We may be better off with a more specific and tightly scoped `team.getSettings(teamSlug: string): TeamSettings` method 👍

DafyddLlyr (commented):
A few comments on syntax and structure 👍 Happy to talk through at any point if unclear.

src/types/team.ts
```diff
@@ -19,7 +19,7 @@ export interface TeamTheme {
   favicon: string | null;
 }
 
-export interface TeamSettings {
+export interface OldTeamSettings {
```
DafyddLlyr, line 22: You should be able to just remove this type here I believe 👍

DafyddLlyr (approved):

PR #436, opened 2024-07-04
feat: get and export original enum values from json schema spec 

jessicamcinchak:
**Problem:**
- We define "enums" in the Digital Planning Data Schemas repository that are converted into a TypeScript type that then generates the schema.json file
- We import that schema.json file in this repository, and regenerate types based on it, but it's difficult to get back to the plain original dictionary enum variables. 
  - We want to access those enum variables, not the types, to do simple validation checks on publish in the Planx Editor (eg "is every `fn` set by a FileUpload or FileUploadAndLabel component supported in the `FileType` definition of the ODP Schema?")

**Changes:**
- Define and make available a few simple utility methods to parse the `schema.json` file and return either a list of the passport values associated with a definition, or the original value/description dictionary
- Basic tests

### DIFF
--- a/src/utils/digitalPlanningSchema.test.ts
+++ b/src/utils/digitalPlanningSchema.test.ts
@@ -1,19 +1,38 @@
-import { getValidSchemaValues } from "./digitalPlanningSchema";
+import {
+  getValidSchemaDictionary,
+  getValidSchemaValues,
+} from "./digitalPlanningSchema";
 
 describe("get valid schema values", () => {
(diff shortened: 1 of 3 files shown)


jessicamcinchak:
[Validate that flow meets schema requirements on "Publish"](https://trello.com/c/LGIoRgym/2641-validate-that-flow-meets-schema-requirements-on-publish)

DafyddLlyr (approved):
Very neat and tidy! Tests look good here 👍

PR #445, opened 2024-07-11
chore: update `formatRawProjectTypes` to reference schema instead of static database table

jessicamcinchak:
The `project_types` database table is very static and this has been causing complaints of 'not found' project types in Notify email templates. Checking against the ODP Schema "ProjectType" definition should ensure a much more up-to-date list and fewer unknowns

**Changes:**
- Updates the `formatRawProjectTypes` function to reference the schema instead of making a GraphQL query
- Moves out of requests and into utils since it is no longer async nor requires a permission-scoped `$client`
- Adds tests
- Adds a polyfill for `Intl.ListFormat` via https://formatjs.io/docs/polyfills/intl-listformat/ as we see this popup occassionally via Airbrake !

**Future planx-new PRs:**
- Update function calls https://github.com/theopensystemslab/planx-new/pull/3410
- Drop `project_types` database table


jessicamcinchak:
[[Research] Can we have configurable templates for GovNotify?](https://trello.com/c/dKhIEOIV/2920-research-can-we-have-configurable-templates-for-govnotify)

src/utils/projectTypes.ts
```diff
@@ -0,0 +1,28 @@
...
+  } else {
+    return "Unknown";
+  }
+}
+
+export function formatHumanReadableProjectTypes(
+  projectTypes: string[],
+): string {
+  // Join in readable format - en-US ensures we use Oxford commas
+  const formatter = new Intl.ListFormat("en-US", { type: "conjunction" });
```
DafyddLlyr, line 22: It would be nice to remove or polyfill this whilst we're touching this function 🙏

jessicamcinchak: Good call - will pick up !

jessicamcinchak: Now updated using this solution: https://formatjs.io/docs/polyfills/intl-listformat/

package.json
```diff
@@ -48,7 +48,6 @@
     "@emotion/react": "^11.11.4",
     "@emotion/styled": "^11.11.5",
     "@mui/material": "^5.15.2",
-    "@types/geojson": "^7946.0.14",
```
jessicamcinchak, line 51: Quickly moved this `@types/` package to the correct spot (`devDependencies`, not regular `dependencies`) while I was here ! Otherwise unrelated to this change.

src/utils/projectTypes.ts
```diff
@@ -0,0 +1,31 @@
...
+  if (schemaProjectTypes) {
+    const matchingDescriptions = Object.entries(schemaProjectTypes)
+      .filter(([key, _description]) => rawProjectTypes.includes(key))
+      .map(([_key, description]) => description.toLocaleLowerCase());
+    if (matchingDescriptions.length === 0) {
+      return "Unknown";
+    }
+    return formatHumanReadableProjectTypes(matchingDescriptions);
+  } else {
+    return "Unknown";
```
jessicamcinchak, line 17: Projects throughout email templates will now fall into one of three cases: 
1) "Project type not submitted" if the passport does not contain `proposal.projectType`
2) List of human-readable project types based on schema if passport does contain `proposal.projectType`
3) "Unknown" if passport does contain `proposal.projectType` but none of those values are valid schema values

I think we should leave "Unknown" as-is for this first refactor and see if/how it shifts feedback. If we're still receiving questions or concerns, there's a few ideas for what we could differently:
- Only return raw project types that are a single segment (eg `breach` &rarr; `Breach`)
- Return all raw project types instead of "Unknown" (either as dot-separated raw variables or perhaps replacing `.` with space and applying formatting?)
- Improve Airbrake logging for these cases (in most cases they are going to be values we _do not_ want to add to schema though, so hesitant to introduce this noise now)

package.json
DafyddLlyr: Good catch!

src/utils/projectTypes.ts
DafyddLlyr: This is a solid plan 👍

package.json
```diff
@@ -69,6 +68,8 @@
     "zod": "^3.23.8"
   },
   "devDependencies": {
+    "@formatjs/intl-listformat": "^7.5.7",
```
DafyddLlyr, line 71: I'm not sure this should be a dev dependency? I think it needs to be a dependency as it's needed in the built version?

DafyddLlyr (approved):
One small question!

package.json
jessicamcinchak: Ah you're totally right - thanks for catching!

PR #462, opened 2024-07-22
chore: Update `IndexedNode` to always have a `parentId` property

DafyddLlyr:
## What?
 - Adds a required `parentId` property to `IndexedNode`
 - Updates `sortFlow()` logic to match PlanX graph structure

## Why?
 - Implemented in https://github.com/theopensystemslab/planx-new/pull/3435
 - For navigation to a node to work from search, we need to know the "path" through the graph to a node
 - Currently `sortFlow()` returns an optional `parentId`, despite all nodes having a parent (apart from the root node). It's more accurate to give all nodes a parent. This makes things a bit easier when constructing a path (upcoming PR)
 - The current implementation of `sortFlow()` doesn't quite match our graph structure. When reaching the next child of the root node, the current parentId is maintained instead of falling back to the root. This worked well for treating the graph as a linear structure to split it up be portal names (pseudo sections), but doesn't work in terms of actually describing the hierarchy of the graph (required to build a valid PlanX URL).
 - **The difference really comes down to "previous node" vs. "parent node"**
- The flow will still retain the same sorted order, `root.edges[0]` → `root.edges[1]` → `root.edges[2]` → etc
 
Note: This should not impact any current functionality. `sortFlow()` and `findNextNodeOfType()` are not used anywhere in PlanX currently.

### Example
 - I have a root node with two edges
 - I navigate through the branches of the first edges until there are no more edges
 - My next step will be the second edge of the root node
 - **Currently:** The parentId for `root.edges[1]` will be set to the very final node of `root.edges[0]` (previous node)
 - **This PR:** The parentId for `root.edges[1]` will be set to "_root" (parent node)


src/models/session/logic.ts
```diff
@@ -107,41 +107,6 @@ export function sortBreadcrumbs(
   return orderedBreadcrumbs;
 }
 
-export function findNextNodeOfType({
```
DafyddLlyr, line 110: This function (and associated tests) has been removed as it's not used, and doesn't work as expected.

jessicamcinchak (approved):

PR #463, opened 2024-07-24
feat: `getPathForFlow()`

DafyddLlyr:
## What does this PR do?
- Generates a path for a node, which is required for search
- Relies on https://github.com/theopensystemslab/planx-core/pull/462
- Implemented in https://github.com/theopensystemslab/planx-new/pull/3460


jessicamcinchak (approved):
Thanks for clear tests!

PR #464, opened 2024-07-25
[templates] extract DrawBoundaryUserAction to src/types

freemvmt:
Small change to fix an import issue in [planx-new](https://github.com/theopensystemslab/planx-new) that appears when trying to bump to ESM.


jessicamcinchak (approved):
Makes sense, thanks for tidying !

PR #465, opened 2024-07-27
bump mui/material and add mui/base as direct dependency

freemvmt:
[@mui/material](https://www.npmjs.com/package/@mui/material?activeTab=dependencies) does not list [@mui/base](https://www.npmjs.com/package/@mui/base) as a dependency, and yet it does rely on it! I think this is probably because base appears to be in alpha, and seems to be directly replacing what was called `@mui/core`.

This was leading to errors like the below when trying to run tests for changes to `planx-new/api.planx.uk` in #[3464](https://github.com/theopensystemslab/planx-new/pull/3464).

```
Cannot find module '@mui/base/composeClasses' from 'node_modules/.pnpm/github.com+theopensystemslab+planx-core@a429c06/node_modules/@opensystemslab/planx-core/node_modules/@mui/material/node/Grid/Grid.js'
```

But explicitly adding `@mui/base` as a dependency in this project (got that idea [here](https://pnpm.io/faq#solution-2)) seems to solve this issue on local. I also bumped `@mui/material` to latest minor version while I was at it :)

**Resources**

- https://pnpm.io/symlinked-node-modules-structure


jessicamcinchak (approved):
Aha great find with this one :bulb: same bug has been plaguing our nightly regression tests in planx-new ! eg https://github.com/theopensystemslab/planx-new/actions/runs/10138961762/job/28031472035

Will go ahead and merge this one today & bump accordingly in planx-new :+1:

PR #479, opened 2024-08-06
feat: Adding domain to the getTeamBySlug query

RODO94:
## What does this PR do?

To enable work being carried out in planx-new: https://github.com/theopensystemslab/planx-new/pull/3483

I want to retrieve the `domain` when fetching a team from the database, so it can be shown to the user in the new Links dialog box

### DIFF
--- a/src/requests/team.ts
+++ b/src/requests/team.ts
@@ -244,6 +244,7 @@ async function getBySlug(client: GraphQLClient, slug: string) {
           id
           name
           slug
+          domain
           theme {
             primaryColour: primary_colour
             actionColour: action_colour


DafyddLlyr (approved):
Approved as this query now matches the `Team` type which is exactly what we want 👍 

However, I'm not totally sure this is what you're looking for in isolation in order to get this in the Editor. There's a function called `initTeamStore()` which makes a query to fetch a team when on any `/:team` route which I think is where we'll need to access `domain`.

This query is a bit of a legacy from before planx-core so it might be nice to replace this with `TeamClient.getBySlug()`? (We'll need to make sure types match - I think the Editor also needs `team_integrations`).

If this isn't clear, happy to take a look and explain myself better on a call tomorrow! Just let me know 👌

RODO94:
@DafyddLlyr Thanks for the comment, I had a suspicion about that function but led with a planx-core first approach...! I'll still merge since it aligns with the `Team` type

PR #487, opened 2024-08-19
chore: Update upsertMember type to generic Role

jamdelion:
## TLDR:
- Update upsertMember type interface to allow any `Role` rather than just `TeamRole`s.
- Add `TeamMember` type here so that it lives in planx-core along with other related types, rather than floating freely in planx-new 🐟 🌊 

## More context:

 Trello ticket: https://trello.com/c/piMcGsJa/2973-add-user-in-editor

I'm working on improving the 'optimistic update' part of my [PR on adding a new team editor](https://github.com/theopensystemslab/planx-new/pull/3500). As per @DafyddLlyr's suggestion, I'm trying to do this by storing the list of team members in the store, rather than [passing as a prop in the route like we do currently](https://github.com/theopensystemslab/planx-new/blob/e0bb2a95f89a72236df3800817bceaa72fcc8b83/editor.planx.uk/src/routes/teamMembers.tsx#L75).

I've done this by adding an `updateTeamMembers` setter to the `TeamStore` in [team.ts](https://github.com/theopensystemslab/planx-new/blob/e0bb2a95f89a72236df3800817bceaa72fcc8b83/editor.planx.uk/src/pages/FlowEditor/lib/store/team.ts#L31):
```
updateTeamMembers: async (teamMembers: TeamMember[]) => {
    const { teamId, $client } = get();
    for (const member of teamMembers) {
      const teamRole = member.role as TeamRole;
      await $client.team.addMember({
        userId: member.id,
        teamId,
        role: teamRole,
      });
    }
  },
  ```
However, this breaks when testing locally because the member `role` is not always a `TeamRole` ([see definition here](https://github.com/theopensystemslab/planx-core/blob/cd5d08de66d5d1f9a7ddeec8700b34cd9220b97b/src/types/roles.ts#L9)) - e.g. there could be a platformAdmin on the team.

Let me know if there is any good reason why I should not change the `upsertMember` interface to allow the more generic `Role` to fix this!

### DIFF
--- a/src/requests/team.ts
+++ b/src/requests/team.ts
@@ -1,14 +1,14 @@
 import type { GraphQLClient } from "graphql-request";
 import { gql } from "graphql-request";
 
-import { TeamRole } from "../types/roles";
+import { Role } from "../types/roles";
 import { Team, TeamSettings, TeamTheme } from "../types/team";
 import { decrypt } from "../utils/encryption";
 
 interface UpsertMember {
   userId: number;
   teamId: number;
-  role: TeamRole;
+  role: Role;
 }
 
 interface RemoveMember {
--- a/src/requests/team.ts
+++ b/src/requests/team.ts
@@ -1,14 +1,14 @@
 import type { GraphQLClient } from "graphql-request";
 import { gql } from "graphql-request";
 
-import { TeamRole, UserRole } from "../types";
+import { EditorRole } from "../types";
 import { Team, TeamSettings, TeamTheme } from "../types/team";
 import { decrypt } from "../utils/encryption";
 
 interface UpsertMember {
   userId: number;
   teamId: number;
-  role: TeamRole | UserRole;
+  role: EditorRole;
 }
 
 interface RemoveMember {
--- a/src/requests/team.ts
+++ b/src/requests/team.ts
@@ -1,14 +1,14 @@
 import type { GraphQLClient } from "graphql-request";
 import { gql } from "graphql-request";
 
-import { Role } from "../types/roles";
+import { TeamRole, UserRole } from "../types";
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/requests/team.ts [hunk 1]
```diff
@@ -1,14 +1,14 @@
...
-import { TeamRole } from "../types/roles";
+import { Role } from "../types/roles";
 import { Team, TeamSettings, TeamTheme } from "../types/team";
 import { decrypt } from "../utils/encryption";
 
 interface UpsertMember {
   userId: number;
   teamId: number;
-  role: TeamRole;
+  role: Role;
```
jessicamcinchak, line 11: I think we might actually want this to be `role: TeamRole | UserRole` which would still get you `platformAdmin`, while restricting/omitting `HasuraRole` (we never want to upsert a member with role "public" !) ? 

Does that make sense? 

Trying to think of correct name for that grouping of roles so could re-use below too, but coming up blank - `TeamRoleWithAdmin` ?? :sweat_smile:

jessicamcinchak (commented):
Thanks for really clear write-up & context - one question though!

src/requests/team.ts
jamdelion: Ah yes that totally makes sense!

Naming wise, maybe something in opposition to `HasuraRole`....`EditorRole`?

jessicamcinchak: Perfect - I like `EditorRole` :+1:

jamdelion:
After discussing with @DafyddLlyr, I realised this PR wasn't needed for what I was trying to do. It may need to be revisited if we extend the 'add a new editor' feature to include platformAdmin users.

PR #489, opened 2024-08-23
feat: adjust queries and types for new `submission_email` DB location

RODO94:
## What does this PR do?

I am adjusting team types and queries to suit the new location of `submission_email` created here: https://github.com/theopensystemslab/planx-new/pull/3554

This is to complete this Trello Ticket: https://trello.com/c/dVsTQpZZ/2984-allow-editors-to-see-update-their-teams-submission-email-in-team-settings-form

This adjustment should help the updating and maintaining of `submission_email` by teamAdmins.

### DIFF
--- a/src/requests/team.ts
+++ b/src/requests/team.ts
@@ -101,7 +101,7 @@ export async function createTeam(
         $name: String!
         $slug: String!
         $domain: String
-        submissionEmail: String
+        $submissionEmail: String
         $settings: team_settings_insert_input!
         $theme: team_themes_insert_input!
       ) {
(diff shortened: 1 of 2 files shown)


src/requests/team.ts
```diff
@@ -111,7 +109,6 @@ export async function createTeam(
             name: $name
             slug: $slug
             domain: $domain
-            submission_email: $submissionEmail
```
jessicamcinchak, line 114: nit: I think this will break e2e tests when merged to planx-new - do you want to wait to _remove_ these fields until the current DB column is also ready to be dropped? 

I think only _additive_ changes in this first type update might be easier to manage and better match the additive-only change you've pushed to planx-new data models?

RODO94: Good shout! I'll revert this change and keep it additive-only.

jessicamcinchak (approved):

PR #500, opened 2024-09-05
refactor: submission email tidy up

RODO94:
## What does this PR do?

I am tidying up work done on submission email, where it was moved from the ``teams`` table to the `team_settings`` table

Changes focused on removing old fields and switching from ``NotifyPersonalisation`` in the type to ``TeamContactSettings`` and using `Pick` to pull them from the ``TeamSettings`` type. This felt right and in line with other changes.

### DIFF
--- a/src/requests/team.ts
+++ b/src/requests/team.ts
@@ -23,7 +23,6 @@ interface NewTeam {
   slug: string;
   domain?: string;
   reference?: string;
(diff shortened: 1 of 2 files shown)


jessicamcinchak (approved):
Appreciate the renaming here - makes more sense based on latest database structure!

PR #507, opened 2024-09-11
fix: remove sub-email from getByTeamSlug

RODO94:
This was added to try and fix a bug with moveFlow, but it didn't fix in the way expected.

Currently fixed with another solution but may revisit for longer term fix.


jessicamcinchak (approved):

jessicamcinchak:
I think it's ideal to only return `submission_email` in cases where we absolutley have to, so if it's possible to safely remove from this generally used `getTeamBySlug` request I'm very happy with this way forward :+1:

PR #509, opened 2024-09-12
feat: new GetTeamSettings query

RODO94:
Adding new query to retrieve TeamSettings only.

In response to `GetBySlug` changing to not return `submission_email` due to column permission changes

### DIFF
--- a/src/requests/team.ts
+++ b/src/requests/team.ts
@@ -366,7 +366,7 @@ async function getIntegrations({
   return decryptedIntegrations;
 }
 
-export async function GetTeamSettings(client: GraphQLClient, slug: string) {
+export async function getTeamSettings(client: GraphQLClient, slug: string) {
   const response: { teams: TeamSettings[] } = await client.request(
     gql`
       query GetTeamSettings($slug: String!) {
--- a/src/requests/team.ts
+++ b/src/requests/team.ts
@@ -374,7 +374,7 @@ export async function getTeamSettings(client: GraphQLClient, slug: string) {
   const response: { teams: TeamSettings[] } = await client.request(
     gql`
       query GetTeamSettings($slug: String!) {
-        team_settings(where: { slug: { _eq: $slug } }) {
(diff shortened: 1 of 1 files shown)


src/requests/team.ts
```diff
@@ -76,6 +76,10 @@ export class TeamClient {
     return getIntegrations({ client: this.client, ...args });
   }
 
+  async getTeamSettings(slug: string): Promise<TeamSettings> {
+    return getTeamSettings(this.client, slug);
+  }
```
DafyddLlyr, line 81: I don't think these types and return values are correctly handled.

This function looks like it actually returns `{ settings: TeamSettings }`, but I think we want to just return a single instance of `TeamSettings` (as per your function type signature here).

DafyddLlyr (changes requested):

src/requests/team.ts
```diff
@@ -366,6 +370,30 @@ async function getIntegrations({
   return decryptedIntegrations;
 }
 
+export async function getTeamSettings(client: GraphQLClient, slug: string) {
+  const response: { team_settings: TeamSettings[] } = await client.request(
+    gql`
+      query GetTeamSettings($slug: String!) {
+        team_settings(where: { team: { slug: { _eq: $slug } } }) {
```
DafyddLlyr, line 377:
```suggestion
  const response: { teamSettings: TeamSettings[] } = await client.request(
    gql`
      query GetTeamSettings($slug: String!) {
        teamSettings: team_settings(where: { team: { slug: { _eq: $slug } } }) {
```

Wherever possible we want to keep all out TypeScript using camelCase and do our casting / destructing from snake_case within the GraphQL query.

DafyddLlyr (approved):
One tiny nitpick before merging please!

src/requests/team.ts
RODO94: Thanks for picking this up!

PR #514, opened 2024-09-16
fix: Get Team Settings response fix

RODO94:
In my last commit for the previous PR, we changed the query so it pulled directly from the `team_settings` table rather than the `teams` table:

```diff
  const response: { teamSettings: TeamSettings[] } = await client.request(
    gql`
      query GetTeamSettings($slug: String!) {
++      team_settings(where: { team: { slug: { _eq: $slug } } }) {
--       teams (where: { slug: { _eq: $slug } } }) {
--       teamSettings: team_settings{
          boundaryUrl: boundary_url
          ...
          submissionEmail: submission_email
        }
      }
    `,
    { slug },
  );
```

But, when I made this change, I forgot to redefine the table to a proper casing `teamSettings: team_settings` which led to issues in what this was returning as the return is:

```ts
  return response.teamSettings[0];
```

### DIFF
--- a/src/requests/team.ts
+++ b/src/requests/team.ts
@@ -391,7 +391,8 @@ export async function getTeamSettings(client: GraphQLClient, slug: string) {
     `,
     { slug },
   );
-  return response.teamSettings[0];
+
+  return response;
 }
 
 async function updateTheme(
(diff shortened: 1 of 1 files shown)


DafyddLlyr (approved):
Thanks for the quick fix here! 🫶 

Apologies for missing this in review, especially after talking about casting snake_case to camelCase 🤦

PR #516, opened 2024-09-19
refactor: alter deleteUser return

RODO94:


jessicamcinchak (commented):
I think there might be a misunderstanding here: queries & mutations in GraphQL that use `_by_pk` suffix will always return _exactly one_ item (based on primary key of table) rather than a list, so the singular `user` is semantically correct here?

Does that make sense? Am I missing something?

jessicamcinchak (approved):
Sorry about prior review, totally me misreading while multitasking :see_no_evil: 

You're absolutely right here, thanks for fixing :+1:

PR #524, opened 2024-10-01
fix: use fallback value for `applicant.siteContact.role` passport mappings

jessicamcinchak:
Follows on from #515 & this latest submission error https://opensystemslab.slack.com/archives/C4B0CKQ3U/p1727787479763779


DafyddLlyr (approved):
Thanks very much for picking this one up 🙌  

It looks like this is also the cause of the failing E2E regression tests. Are you ok to please also open a PR to bump across planx-new?

image

PR #534, opened 2024-10-14
fix: ensure Gov Pay reference is set for invite to pay applications in overview HTML and CSV

jessicamcinchak:
Fixes #help-issues report from Friday afternoon: https://opendigitalplanning.slack.com/archives/C0241GWFG4B/p1728640257349149

Depending if the applicant self-paid or invited-to-pay, the payment info will be stored differently. We've been correctly accounting for this in the ODP Schema ([eg here](https://github.com/theopensystemslab/planx-core/blob/main/src/export/digitalPlanning/model.ts#L686-L698)) but NOT the CSV/Overview HTML formats and the reference number in invite to pay cases was failing to populate.

This fix correctly accounts for invite to pay cases, but ideally in the near future we'll do a larger refactor to generate the CSV & Overview HTML from the ODP Schema directly so we can remove these duplicated bespoke messy mappings!

### DIFF
--- a/src/export/csv/index.ts
+++ b/src/export/csv/index.ts
@@ -3,6 +3,7 @@ import { omit } from "lodash";
 import { Passport } from "../../models";
 import type {
   BOPSFullPayload,
+  GovUKPayment,
   Passport as IPassport,
   QuestionAndResponses,
   Response,
@@ -12,10 +13,12 @@ export function computeCSVData({
   sessionId,
   bopsData,
   passport,
+  govUkPayment,
 }: {
   sessionId: string;
   bopsData: BOPSFullPayload;
   passport: IPassport;
+  govUkPayment?: GovUKPayment;
 }): QuestionAndResponses[] {
   // format dedicated BOPS properties (eg `applicant_email`) as list of questions & responses to match proposal_details
   //   omitting debug data and payload keys already in confirmation details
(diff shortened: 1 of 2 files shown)


jessicamcinchak:
[Ensure overview.html & application.csv documents correctly populate Gov Pay reference for invite to pay submissions](https://trello.com/c/liYUE24Y/3087-ensure-overviewhtml-applicationcsv-documents-correctly-populate-gov-pay-reference-for-invite-to-pay-submissions)

src/export/csv/index.ts
```diff
@@ -68,13 +71,21 @@ export function computeCSVData({
...
-  ];
+
+  // check if the passport has payment info (or session govUkPayment if invite to pay), add as reference row if exists
+  const selfPayPassportKey = "application.fee.reference.govPay";
+  if (passportInstance.has([selfPayPassportKey]) || govUkPayment) {
+    references.push({
+      question: selfPayPassportKey,
+      responses:
+        (passportInstance.any([selfPayPassportKey]) as Response[]) ||
+        (govUkPayment as unknown as Response[]),
```
jessicamcinchak, line 83: I know this is ugly type casting, but does what we need with minimal cascading changes to these bits of old tedious code !

DafyddLlyr (approved):
Thanks for picking this one up - totally agreed that the aim here should be to generate this from the schema format 👌

PR #541, opened 2024-10-25
chore: Error handling for `decrypt()`

DafyddLlyr:
## What does this PR do?
- Wraps `decrypt()` in a try/catch
- Currently, if decryption fails, an unhandled exception brings down Express
- This now ensures it's handled gracefully, and the consumer needs to check the existence of they keys (the function still maintains it's signature of returning `string | undefined` for a secret)

## To test 
 - Make a secret invalid, e.g. populate encrypted GovPay token as plaintext string `"abc123"`
 - Try to make payment
 - Graceful error handling in application, Docker logs capture issue
 - Example flow here on Pizza with an invalid GovPay key - https://3848.planx.pizza/epsom-and-ewell/test-pay-failure/published?analytics=false

image

image

image

### DIFF
--- a/src/utils/encryption.ts
+++ b/src/utils/encryption.ts
@@ -31,14 +31,19 @@ export function decrypt(
 ): string | undefined {
   if (!secret) return undefined;
 
-  const [encryptedToken, iv] = secret.split(":");
-  const decipher = crypto.createDecipheriv(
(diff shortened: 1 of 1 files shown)


jessicamcinchak (approved):
Perfect thanks !

PR #553, opened 2024-11-05
fix: Correctly track portal IDs when indexing nodes

DafyddLlyr:
## What does this PR do?
 - Fixes issue described here https://opensystemslab.slack.com/archives/C01E3AC0C03/p1730805531747509 (OSL Slack)
 - Creates a stack of portal ids when indexing nodes
   - Add to the stack when we hit a portal
   - Remove from the stack when we exit
 - This means that we now correctly assign the `internalPortalId` to all indexed nodes.

Implemented in https://github.com/theopensystemslab/planx-new/pull/3914

### DIFF
--- a/src/models/session/logic.ts
+++ b/src/models/session/logic.ts
@@ -14,10 +14,9 @@ import { ComponentType } from "../../types";
 
 export function sortFlow(flow: FlowGraph): OrderedFlow {
   let sectionId: string | undefined;
-  let internalPortalId: string | undefined;
-
   const nodes: IndexedNode[] = [];
   const nodeIds = new Set<string>();
+  const internalPortals: string[] = [];
 
   const searchNodeEdges = (id: string, parentId: string) => {
     // Skip already added nodes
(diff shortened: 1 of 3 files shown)


src/models/session/logic.ts
```diff
@@ -28,32 +27,36 @@ export function sortFlow(flow: FlowGraph): OrderedFlow {
...
+    // If we've hit a portal, add it to the stack
+    if (type === ComponentType.InternalPortal) internalPortals.push(id);
 
-    foundNode.edges?.forEach((childEdgeId) => {
+    edges?.forEach((childEdgeId) => {
       searchNodeEdges(childEdgeId, id);
     });
+
+    // Remove portal from stack when we exit it
+    if (type === ComponentType.InternalPortal) internalPortals.pop();
```
jessicamcinchak, line 59: :+1: tricky but makes perfect sense !

src/models/session/mocks/flow-with-internal-portals.ts
```diff
@@ -0,0 +1,104 @@
...
+  DJDsTYVeZG: {
+    data: {
+      fn: "animal",
+      tags: [],
+      text: "Elephant",
+      neverAutoAnswer: false,
+    },
+    type: ComponentType.Question,
+    edges: ["LNjFPmokA4"],
+  },
```
jessicamcinchak, line 16: nit: appreciate that you've updated to human-readable `type` here, but still a bit hard to read tests referencing random node IDs - eg `DJDsTYVeZG` &rarr; `levelOneQuestionTwo` would help me ! (Even if just for node IDs that show up in `expect()` statements, no need to rename `Answer` IDs necessarily ! )

jessicamcinchak (approved):
Thank for getting to the bottom of this one!

src/models/session/mocks/flow-with-internal-portals.ts
DafyddLlyr: No this is a very fair and good comment thanks - changed in [145a49b](https://github.com/theopensystemslab/planx-core/pull/553/commits/145a49b23a9238c0c6cec70b2dc4bddec8857a97)

PR #554, opened 2024-11-06
fix: filter questions without responses out of generated docs

jessicamcinchak:
Per bug report here https://opensystemslab.slack.com/archives/C01E3AC0C03/p1730907777602279

Questions without `edges` (aka "sticky note" questions) now leave a breadcrumb per the new automation logic, whereas before they didn't. They were being correctly filtered out of the Result & Review pages in-service, but incorrectly included in the submission documents.

(These areas of code are in need of a BIG need of refactor/simplification, but purely fixing-forward/patching right now! Sorry for messy commits too, was having a bunch of cache issues testing locally in-planx-new - but all good on pizza :see_no_evil:)

### DIFF
--- a/src/templates/html/application/ApplicationHTML.tsx
+++ b/src/templates/html/application/ApplicationHTML.tsx
@@ -178,10 +178,13 @@ function SectionList(props: { data: PlanXExportData[] }) {
 }
 
 function DataItem(props: { data: PlanXExportData }) {
+  // Filter out any items that don't have a response (eg "sticky note" questions)
   if (
-    ["Error displaying response", ""].includes(
-      prettyResponse(props.data.responses),
-    )
+    [
+      "Error displaying response",
+      "Error displaying list of responses",
+      "",
+    ].includes(prettyResponse(props.data.responses))
   )
     return;
 


jessicamcinchak:
[Automation & filtering logic followups](https://trello.com/c/dboRkua3/3112-automation-filtering-logic-followups)

DafyddLlyr (commented):
Agreed that there's a need to revisit some of this code and tidy it up! Been back in it with https://github.com/theopensystemslab/planx-core/pull/555 today (long long overdue!).

I _think_ there's an easier solution here where we just stop generating responses if there are no answers for the node - 

```ts
// formatProposalDetails()

...
    default:
      return crumb.answers; // Drop empty array
  }
})()

if (!answers) continue; // Quit loop if no answers for node
```

I've not tested this out but I think it will work. 

If it doesn't, and we need a more manual approach we also need to update the CSV generation which still relies on `computeBOPSParams()`.

jessicamcinchak:
Thanks @DafyddLlyr ! That suggestion works as expected and is much tidier :star2: 

Closing this one in favor of #556

PR #555, opened 2024-11-06
feat: Generate `QuestionAndResponse[]` for `List` and `Page` components

DafyddLlyr:
## What does this PR do?
 - Ensures that when a payload contains a `List` or `Page` component, that the values are parsed
 - These parsed values are used in the following ways - 
   - Responses in `DigitalPlanningPayload`
   - Responses in CSV
   - Responses in generated HTML docs

There's a lot of room for refactors and improvements in this area of the code base, but I've tried to make this an additive change here instead of starting to untangle too much at the same time.

### DIFF
--- a/src/export/bops/index.ts
+++ b/src/export/bops/index.ts
@@ -29,6 +29,7 @@ import {
   USER_ROLES,
 } from "../../types";
 import { DataObject } from "./../../types/data";
+import { getSchemaProposalDetails, isSchemaType } from "./schema/utils";
 
 const bopsDictionary = {
   // applicant or agent details can be provided via TextInput(plural) or ContactInput component
@@ -51,11 +52,13 @@ function isTypeForBopsPayload(type?: ComponentType) {
   //To-Do Set MapAndLabel to True once we know shape of BOPs data
 
   switch (type) {
+    case ComponentType.Answer:
     case ComponentType.Calculate:
     case ComponentType.Confirmation:
     case ComponentType.Content:
     case ComponentType.DrawBoundary:
     case ComponentType.ExternalPortal:
+    case ComponentType.Feedback:
     case ComponentType.FileUpload:
     case ComponentType.FileUploadAndLabel:
     case ComponentType.Filter:
@@ -67,26 +70,24 @@ function isTypeForBopsPayload(type?: ComponentType) {
     case ComponentType.Pay:
     case ComponentType.PlanningConstraints:
     case ComponentType.PropertyInformation:
-    case ComponentType.Answer:
     case ComponentType.Result:
     case ComponentType.Review:
     case ComponentType.Section:
     case ComponentType.Send:
     case ComponentType.SetValue:
     case ComponentType.TaskList:
-    case ComponentType.List:
-    case ComponentType.MapAndLabel:
-    case ComponentType.Page:
-    case ComponentType.Feedback:
       return false;
 
     case ComponentType.AddressInput:
     case ComponentType.Checklist:
+    case ComponentType.ContactInput:
     case ComponentType.DateInput:
+    case ComponentType.List:
+    case ComponentType.MapAndLabel:
     case ComponentType.NumberInput:
+    case ComponentType.Page:
     case ComponentType.Question:
     case ComponentType.TextInput:
-    case ComponentType.ContactInput:
       return true;
 
     default:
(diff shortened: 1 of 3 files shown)


src/export/bops/index.ts
```diff
@@ -48,45 +49,43 @@ const bopsDictionary = {
...
     case ComponentType.DrawBoundary:
     case ComponentType.ExternalPortal:
+    case ComponentType.Feedback: // TODO
     case ComponentType.FileUpload:
     case ComponentType.FileUploadAndLabel:
     case ComponentType.Filter:
     case ComponentType.FindProperty:
     case ComponentType.Flow:
     case ComponentType.InternalPortal:
+    case ComponentType.MapAndLabel: // TODO - Parse GeoJSON FeatureCollection
```
DafyddLlyr, line 66: I was also going to pick this up here, but the data structure it outputs is actually different. This will be a follow up PR shortly!

src/export/bops/index.ts
```diff
@@ -159,6 +158,29 @@ export function formatProposalDetails({
...
+
+    if (isSchemaType(crumb.type)) {
+      const schemaProposalDetails = getSchemaProposalDetails(
+        crumb,
+        node,
+        metadata,
+      );
+      proposalDetails.push(...schemaProposalDetails);
+      continue;
+    }
```
DafyddLlyr, line 182: Untouched block, just moved higher in the function so that we can pass `metadata` along.

src/export/bops/index.ts
```diff
@@ -159,6 +158,29 @@ export function formatProposalDetails({
...
+      if (hasSections) {
+        metadata.section_name = currentSectionName;
+      }
+      if (crumb.autoAnswered) metadata.auto_answered = true;
+      if (node.data?.policyRef) {
+        const policyRefs = parsePolicyRefs(node.data.policyRef as string);
+        if (policyRefs.length) metadata.policy_refs = policyRefs;
+      }
+      return metadata;
+    })();
```
DafyddLlyr, line 172: Untouched block, just moved higher in the function so that we can pass `metadata` along.

src/export/bops/index.ts
```diff
@@ -159,6 +158,29 @@ export function formatProposalDetails({
...
+
+    if (isSchemaType(crumb.type)) {
```
DafyddLlyr, line 174: **Why is this not handled in the normal per-component pattern?**

Here's how the works currently - 
 - Iterate over answers in breadcrumbs 
 - Find corresponding node in flow, to get the question put to users
 
This pattern doesn't work for more complex, schema driven, components. This is because there's not a 1-to-1 relationship between questions and answers per-node as in the previously established pattern.

We need to break out of this loop to handle schema driven components so that we can create multiple questions and responses per-node.

src/export/bops/schema/utils.ts
```diff
@@ -0,0 +1,140 @@
...
+  Response,
+} from "../../../types";
+
+export const isSchemaType = (type: ComponentType) =>
+  [ComponentType.Page, ComponentType.List].includes(type);
+
+/**
+ * Parse the data of a node which uses a schema for responses
+ */
+const parseNodeData = (node: Node) => {
```
DafyddLlyr, line 17: This could maybe be a bit easier if we had the `Schema` types from the frontend here, but this is a nice way of parsing an `unknown` into the type we expect.

In an ideal world, the Flow and Breadcrumb types would be shared and node-specific, e.g. `EncrichedCrumb<Page>` or `NodeData<List>`.

jessicamcinchak (approved):
Overall looks good, thank for picking this one up !

src/export/bops/tests/proposalDetails.test.ts
```diff
@@ -395,3 +399,1414 @@ describe("Flow without sections", () => {
...
+          "mockData.total.units": 9,
+        },
+      },
+    };
+
+    const expected = {
+      proposalDetails: [
+        {
+          question:
+            "[Residential units (GLA) - Gained] Number of units of this type (Gained residential unit type 1)",
```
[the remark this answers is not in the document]
DafyddLlyr: Nice suggestion from both a users's POV and a code simplification 👍

src/export/bops/index.ts
DafyddLlyr: Please see https://github.com/theopensystemslab/planx-core/pull/562

PR #562, opened 2024-11-11
feat: Generate `QuestionAndResponse[]` for `MapAndLabel` component

DafyddLlyr:
## What does this PR do?
- Follow up to https://github.com/theopensystemslab/planx-core/pull/555, specifically https://github.com/theopensystemslab/planx-core/pull/555#discussion_r1832518335
- Ensures that payloads containing a `MapAndLabel` component are correctly parsed in the `QuestionAndResponse[]` format which is used in many places


jessicamcinchak (approved):
Perfect - thanks again for picking up these long-neglected tasks !

PR #564, opened 2024-11-14
feat: Add `insertFlowDescription` Mutation

RODO94:
## What does this PR do?

This PR sets up work into Templates: https://trello.com/c/YjhjFaIQ/3111-templates-groundwork

Following work to add a `description` column to the `flows` table, I am not setting up work to add functionality to update it in the Editor

### DIFF
--- a/src/requests/flow.ts
+++ b/src/requests/flow.ts
@@ -148,7 +148,6 @@ export async function createFlow(
         $teamId: Int!
         $flowSlug: String!
         $flowName: String!
-        $description: String
         $data: jsonb
         $status: flow_status_enum_enum
       ) {
@@ -160,7 +159,6 @@ export async function createFlow(
             data: $data
             version: 1
             status: $status
-            description: $description
           }
         ) {
           id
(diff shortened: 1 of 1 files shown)


src/requests/flow.ts
```diff
@@ -58,7 +62,7 @@ export async function getLatestFlowGraph(
   const response: { published_flows: { data: FlowGraph }[] } =
     await client.request(
       gql`
-        query GetLatestPublishedFlowData($flowId: uuid!) {
+        query GetLatestPublishedFlowData($flowId: string!) {
```
DafyddLlyr, line 65: Why has this changed?

src/requests/flow.ts
```diff
@@ -167,6 +173,7 @@ export async function createFlow(
       flowName: args.name,
       data: args.data,
       status: args.status || "offline",
+      description: "",
```
DafyddLlyr, line 176: I think it might make more sense to consider this a nullable field - current flows don't have descriptions and I don't think a description will be mandatory for a flow.

src/requests/flow.ts
RODO94: Sorry! Yes, this was a change I made to the wrong query, could pick up, I'll revert

DafyddLlyr (approved):

PR #579, opened 2024-12-05
 feat: Calculate VAT using `getFeeBreakdown()`

DafyddLlyr:
## What does this PR do?
 - Changes logic to now rely on on `application.fee.includesVAT` (boolean) instead of `application.fee.vat` (number)
 - This means we are relying on Editors to tell us if the fee in inclusive of VAT, but does not rely on them to make the calculation themselves
 - Based on decision from Monday's fee breakdown show & tell

### Note
This doesn't cover the cases where teams require a VAT amount to be included within their GovUK Pay metadata. In these cases they will still need to calculate it as a passport variable, and add it to the metadata fields. It's probably worth getting teams to adopt this approach there as well but I'm intentionally avoiding this here in order to focus on the fee breakdown feature.

### DIFF
--- a/src/types/feeBreakdown.ts
+++ b/src/types/feeBreakdown.ts
@@ -12,7 +12,7 @@ export interface FeeBreakdown {
 export interface PassportFeeFields {
   "application.fee.calculated": number;
   "application.fee.payable": number;
-  "application.fee.payable.vat": number;
+  "application.fee.payable.includesVAT": boolean;
   "application.fee.reduction.alternative": boolean;
   "application.fee.reduction.parishCouncil": boolean;
   "application.fee.reduction.sports": boolean;
--- a/src/utils/feeBreakdown.test.ts
+++ b/src/utils/feeBreakdown.test.ts
@@ -27,7 +27,7 @@ describe("calculateReduction() helper function", () => {
     const input: PassportFeeFields = {
       "application.fee.calculated": 100,
       "application.fee.payable": 50,
-      "application.fee.payable.vat": 0,
+      "application.fee.payable.includesVAT": false,
       "application.fee.reduction.alternative": false,
       "application.fee.reduction.parishCouncil": false,
       "application.fee.reduction.sports": false,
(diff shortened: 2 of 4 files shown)


jessicamcinchak (approved):
Looks good ! 

A couple non-blocking questions - 
- If an editor toggles on `includesVAT`, what would be the consequences of then automatically adding a row with the calculated number (eg `application.fee.payable.vat`) to their GOV Pay metadata (row limit, other configured metadata fields?) rather than what you mention in comment of them needing to use SetValue/Calculate prior?
- Arguably, I think we'll also want the _number_ VAT rather than boolean at point of submission for the ODP Schema, can the Pay component's `onSubmit` write `application.fee.payable.VAT` (number type) to the passport if `includesVAT` is true?

DafyddLlyr:
Both great questions 👍 

> If an editor toggles on `includesVAT`, what would be the consequences of then automatically adding a row with the calculated number (eg `application.fee.payable.vat`) to their GOV Pay metadata (row limit, other configured metadata fields?) rather than what you mention in comment of them needing to use SetValue/Calculate prior?

It's a good point to raise - the downside is that all services will lose one of their 10 metadata fields to this, even if VAT is always zero and never applicable to this service. Doncaster are already having issues with the number of available fields and don't think they can mange with 10. It would certainly be a better outcome though in terms of Editor experience (VAT is always tracked and calculated, you just have to set if it applies or not). I'll circle back to this one post fee-breakdown work. Blue trello card to discuss this one here - https://trello.com/c/c3AsVDP7/3174-should-we-include-the-vat-amount-automatically-as-govuk-pay-metadata

> Arguably, I think we'll also want the number VAT rather than boolean at point of submission for the ODP Schema, can the Pay component's `onSubmit` write `application.fee.payable.VAT` (number type) to the passport if `includesVAT` is true?

There's no slot for this in the schema currently, but I agree that this might come up as a requirement - if it does, we'll want the amount not a boolean for sure. I'm hoping that I can use `getFeeBreakdown()` to populate the passport mapping inside `getApplicationFee()` in planx-core, which already handles this VAT calculation.

PR #595, opened 2024-12-19
migrate: proposed flag changes

jessicamcinchak:
Per data migration plan here: https://docs.google.com/spreadsheets/d/1Vtxp5BLweDPDooQoNhgOCYjCIBPRYIcOuyArGJRqOkI/edit?gid=951386129#gid=951386129

Plus conversation here: https://opensystemslab.slack.com/archives/C07F7215W7P/p1734522579596699

Mock data updates are VERY dense here & rely on previously merged new schema release v0.7.2 - these flows have been copied fresh from staging again this morning post-migration scripts, as a manual find+replace isn't realistic. Basically, this acts as another test layer against the scripts themselves though this way !


src/export/bops/mocks/payload.ts
```diff
@@ -1195,36 +1195,13 @@ export const mockExpectedBOPSPayload = {
...
-      question: "Are there any protected trees on the property?",
-      responses: [
-        {
-          value: "No",
-        },
-      ],
-      metadata: {
-        auto_answered: true,
-      },
-    },
```
jessicamcinchak, line 1220: These two questions are no longer in the up-to-date flow, so we no longer expect them to be in the payload either

src/types/flags.ts
```diff
@@ -1,6 +1,6 @@
 export type FlagSet =
   | "Planning permission"
-  | "Listed building consent"
+  | "Works to listed buildings"
```
jessicamcinchak, line 3: This file is the important change! Once merged to planx-new, we should see all of our flag color bands re-appear as well as _new_ Filters added having updated data values & text set here

DafyddLlyr (approved):

PR #601, opened 2024-12-24
chore: Update planx-core to ESM module

DafyddLlyr:
## What does this PR do?
 - Converts this project from CommonJS to ESM
 - Associated changes - 
   - Convert from jest to vitest
   - Convert from ts-node to tsx
   - Convert from lodash to lodash-es
   - Add type assertions to JSON imports
   - Fix imports to use `.js` in filenames

Some of these changes were a bit haphazard - one following the other as I hit issues. If anything is too unclear / too big, please shout and I'll work on splitting this up.

## Why?
 - Required for https://github.com/theopensystemslab/planx-core/pull/600 (import ESM schema lib)
 - Similarly, this might be needed if we decide to publish this repo as a package
 - ESM is the way things are headed - worth making the change and not publishing in CommonJS!


jessicamcinchak (approved):
Thanks for this !! Dense changes indeed but all looks good here - and corresponding planx-new PR/pizza seems healthy :tada:

freemvmt (approved):
Looks great! Brings up happy mems of doing the same for the [API](https://github.com/theopensystemslab/planx-new/pull/3464) 😅

My only question is whether there is some way to simplify the tsconfig setup? e.g. is one an extension of another. and can be represented as such? could the canonical `tsconfig.json` replace any of the particular `tsconfig.x.json` files? haven't looked at it closely, but just a thought!

PR #611, opened 2025-01-07
migrate: find+replace `flag` prop with `flags` across all mock flow data

jessicamcinchak:
Per data migration plan here: https://docs.google.com/spreadsheets/d/1Vtxp5BLweDPDooQoNhgOCYjCIBPRYIcOuyArGJRqOkI/edit?gid=0#gid=0

Corresponds to planx-new PR here: https://github.com/theopensystemslab/planx-new/pull/4098

**Changes:**
- planx-core has multiple instances of very large mock flow data and instead of a find+replace of `flag` to `flags` (plus making value an array) across these, I've simply copied new versions from the entire flows from staging post-database migration scripts this morning. This means they now have the correct structure, but general content changes as well
- `getResultData` methods are updated and synced to planx-new preview store (see comment below)

### DIFF
--- a/src/models/result.ts
+++ b/src/models/result.ts
@@ -2,7 +2,6 @@ import type {
   Breadcrumbs,
   Flag,
   FlagSet,
-  FlagValue,
   FlowGraph,
   ResultData,
 } from "../types";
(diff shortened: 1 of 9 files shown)


src/models/result.ts
```diff
@@ -13,58 +12,63 @@ import {
   resultOverrides,
 } from "../types";
 
+// Functions in this file need to stay in sync with planx-new/editor.planx.uk/src/pages/FlowEditor/lib/store/preview.ts !!
+//   TODO eventually replace/combine ??
```
jessicamcinchak, line 16: This is important and annoying ! I'd missed this when updating result store functions in planx-new around the automation changes - which means the CSV has only been including result data based on the legacy "flag" format of old questions/checklist options until now.

This change brings the functions back in sync, but ultimately we need to revisit this so that we don't need to maintain these in x2 places to begin with. (Can we store `result` on session data & eliminate need to "calculate" it in planx-core at all?)

src/models/session/logic.test.ts
```diff
@@ -122,8 +122,8 @@ describe("getPathForNode", () => {
   it("returns a path for a complex flow", () => {
     const path = getPathForNode({ nodeId: "kTEuqpqCh2", flow });
 
-    expect(path).toHaveLength(58);
-    expect(path[57].id).toBe("_root");
+    expect(path).toHaveLength(62);
+    expect(path[61].id).toBe("_root"); // _root is always last
```
jessicamcinchak, line 126: The mock flows changed here so the length of node paths has legitimately changed too

src/models/result.ts
RODO94: Agree that this would be handly, took me a bit to work out the need for the two. 

Are they both doing the same thing but using the return in different places. We recalculate the `result` for payload generation in `planx-core` while we use the `resultData` in the `Result` component in `planx-new`? ?

RODO94 (approved):
Ran through the changes to check for any typos or overlooked changes but couldn't find any. Skimming the mocks seemed unwieldy so I kept that to a `ctrl+f` for possible typos or missed ones, but I imagine these to be pretty spot on. 

These changes are scheduled before switching the flag values in the `flatFlags` array. Will we have to change the mocks again at this point to incorporate the new values? 

I think it's the right approach to go step by step with the changes, I just want to check my understanding of upcoming changes on this.

All looks good to me though

src/models/result.ts
jessicamcinchak: Correct - result data is a computation based on breadcrumbs & flow data like the passport, but unlike the passport which is saved on `lowcal_sessions.data`, result is computed on the fly in planx-new and therefore needs to be re-computed when referenced in planx-core rather than simply fetched from the session data like we can with a passport

jessicamcinchak:
@RODO94 thanks for review! Yes this change will merge first, then https://github.com/theopensystemslab/planx-core/pull/595 will need to be rebased / updated - but that PR is schema-release-dependent as well so not part of this initial data migration

PR #622, opened 2025-01-13
feat: Map sameSiteAddress variable

DafyddLlyr:


src/export/digitalPlanning/model.ts
```diff
@@ -314,6 +314,28 @@ export class DigitalPlanning {
     );
   }
 
+  private getApplicantAddress = (): Payload["data"]["applicant"]["address"] => {
+    const isSameSiteAddress =
+      (this.passport.data?.[
+        "applicant.address.sameAsSiteAddress"
+      ]?.[0] as string) === "Yes" ||
```
jessicamcinchak, line 321: Based on the latest `opensystemslab/about-the-applicant` I actually think we want to be mapping against the variable `applicant.resident` & possible values `true` or `false` here as the primary check? https://editor.planx.uk/opensystemslab/about-the-applicant,XQsWKqj72N/nodes/D95sd8eRww/nodes/9pjmVfdzHW/edit

Legacy variable matches the SetValue as expected :+1:

jessicamcinchak (approved):
Thanks for picking this one up! Long overdue, apologies this was never correctly marked as a "TODO" in the code

src/export/digitalPlanning/model.ts
DafyddLlyr: That is super helpful thanks - I was working from the older payload here.

PR #625, opened 2025-01-18
chore: bump schema to v0.7.2

jessicamcinchak:
**Changes:**
- Copy+paste latest `application` schema json
- Regenerate types
- Updates mock flows and sessions to validate (merges #623 & updates `property.type` changes onto this branch for passing tests)

**To revisit after initial merge:**
- [ ] Adjust build configs to account fo `application` schema _and_ `preApplication` schema
- [ ] Copy+paste and generate types for `preApplication` schema
- [ ] Basic passport mappings for `preApplication` schema
- [ ] Add mock `preApplication` session for tests

PR #629, opened 2025-01-21
chore: add support for multiple schemas within `export/digitalPlanning`

jessicamcinchak:
Picks up outstanding tasks outlined here https://github.com/theopensystemslab/planx-core/pull/625

See implementation in https://github.com/theopensystemslab/planx-new/pull/4206

We now have a `preApplication` schema in addition to an `application` schema and need to be able to correctly map and validate any PlanX passport against the correct one to generate the `application.json` files included in submission payloads.

The logic is:
- If `application.type = "preApp"`, then make and validate a `preApplication` payload
- If `application.type` is any defined statutory application type, then make and validate an `application` payload
- If `application.type` is unrecognised or undefined, then make an `application` payload with `skipValidation = true`


package.json
```diff
@@ -42,8 +42,8 @@
     "check": "tsc --project ./tsconfig.check.json && pnpm lint",
     "postinstall": "pnpm i rimraf && pnpm build",
     "prepare": "husky install",
-    "generate-types-from-schema": "cd src/export/digitalPlanning/schema/ && json2ts schema.json > types.d.ts",
-    "copy-json-schema": "copyfiles -f ./src/export/digitalPlanning/schema/schema.json ./dist/export/digitalPlanning/schema"
+    "generate-types-from-schema": "cd src/export/digitalPlanning/schemas/ && json2ts application/schema.json > application/types.d.ts && json2ts preApplication/schema.json > preApplication/types.d.ts",
+    "copy-json-schema": "copyfiles -f ./src/export/digitalPlanning/schemas/application/schema.json ./dist/export/digitalPlanning/application/schema && copyfiles -f ./src/export/digitalPlanning/schemas/preApplication.json ./dist/export/digitalPlanning/preApplication/schema"
```
jessicamcinchak, line 46: These commands aren't the most eloquent and will be worth refactoring to crawl the directory or similar if we find ourselves with a third schema, but they work for now (and no known plans for a third schema anytime soon)!

src/export/digitalPlanning/index.ts
```diff
@@ -12,7 +13,9 @@ export async function generateDigitalPlanningPayload({
   client,
   sessionId,
   skipValidation,
-}: DigitalPlanningExportParams): Promise<DigitalPlanningPayload> {
+}: DigitalPlanningExportParams): Promise<
+  ApplicationPayload | PreApplicationPayload
+> {
```
jessicamcinchak, line 18: In an ideal world, I think we'd have two separate classes that each return a precise type rather than a single class that returns `ApplicationPayload | PreApplicationPayload` - but this approach seemed like the path of least resistance to share as many of the existing passport helper methods as possible without a massive refactor or duplication (eg they're `private` functions!)

src/utils/digitalPlanningSchema.ts
```diff
@@ -19,7 +19,7 @@ export function getValidSchemaValues(definition: string): string[] | undefined {
 }
 
 /**
- * For a given 'anyOf' enum definition in the JSON Schema (with properties `value` & `description`), return a dictionary of its valid options
+ * For a given 'anyOf' enum definition in the Application JSON Schema (with properties `value` & `description`), return a dictionary of its valid options
```
jessicamcinchak, line 22: These helper functions (which we call in planx-new for things like publish validation checks) are going to continue only reading against the `application` schema - but I think this is generally fine & expected behavior! Enums referenced throughout `preApplication` are shared anyways currently.

src/export/digitalPlanning/model.ts
```diff
@@ -133,14 +143,60 @@ export class DigitalPlanning {
...
+        proposal: {
+          description: this.passport.data?.["proposal.description"] as string,
+          boundary: this.getProposedBoundary(),
+        },
+      },
+      responses: this.getResponses(),
+      files: this.getFiles(),
+      metadata: this.getMetadata() as PlanXMetadata,
+    };
+  }
```
jessicamcinchak, line 187: Similar to comments above, I'm introducing a new `mapPassportToPreApplicationPayload()` within the _existing_ class here to share/re-use as many of the existing mappings as possible.

`this.payload` and `validatePayload()` are the only other pieces that need to be application-type aware.

package.json
```diff
@@ -42,8 +42,9 @@
     "check": "tsc --project ./tsconfig.check.json && pnpm lint",
     "postinstall": "pnpm i rimraf && pnpm build",
     "prepare": "husky install",
-    "generate-types-from-schema": "cd src/export/digitalPlanning/schema/ && json2ts schema.json > types.d.ts",
-    "copy-json-schema": "copyfiles -f ./src/export/digitalPlanning/schema/schema.json ./dist/export/digitalPlanning/schema"
+    "generate-types-from-schema": "cd src/export/digitalPlanning/schemas/ && json2ts application/schema.json > application/types.d.ts && json2ts preApplication/schema.json > preApplication/types.d.ts",
+    "copy-json-schema": "copyfiles -f ./src/export/digitalPlanning/schemas/application/schema.json ./dist/export/digitalPlanning/schemas/application",
+    "copy-json-preapp-schema": "copyfiles -f ./src/export/digitalPlanning/schemas/preApplication/schema.json ./dist/export/digitalPlanning/schemas/preApplication"
```
jessicamcinchak, line 47: These commands aren't the most eloquent and will be worth refactoring to crawl the directory or similar if we find ourselves with a third schema, but they work for now (and no known plans for a third schema on our side anytime soon)!

jamdelion (approved):
Clear to follow (once I'd unfurled `model.ts` anyway, before that I was a bit confused! 😅 ) and all looks sensible to me! Definitely pro sharing as many methods as possible, and happy with the slight duplication throughout while we just have two schemas - still easy to read.

src/export/digitalPlanning/index.ts
DafyddLlyr: Agreed on general approach here for sure.

If we hit any type issues with this in planx-new in the short term, we could maybe consider passing an additional argument into the class - 

```ts
type DigitalPlanningExportParams = {
  client: Client;
  sessionId: string;
  skipValidation?: boolean;
  type: "preApp" | ApplictionTypeUnion; // Create type discriminator here
};

export async function generateDigitalPlanningPayload({
  client,
  sessionId,
  skipValidation,
  type,
}: DigitalPlanningExportParams): Promise
 //  Modify return type based on this
  type extends "preApp'"? PreApplicationPayload : ApplicationPayload
> {
  ...
}
```

Returning a union though is exactly what I would have done here - I don't think we use the return type much in PlanX at this stage.

src/export/digitalPlanning/model.ts
```diff
@@ -133,14 +143,60 @@ export class DigitalPlanning {
...
+
   validatePayload() {
     const ajv = addFormats.default(new Ajv.default({ allowUnionTypes: true }));
-    const validate = ajv.compile(jsonSchema);
+    const validate =
+      this.applicationType === "preApp"
+        ? ajv.compile(preApplicationJsonSchema)
+        : ajv.compile(applicationJsonSchema);
```
DafyddLlyr, line 194: One thought on the whole multiple classes thing I've just hit seeing some of this code. 

We can inherit based on classes so maybe there's a decent way forward here without a huge rewrite that has the majority of the functionality in a single abstract / base class, and then have per-app type classes that _just_ handle the differences. 

This might actually make things a whole lot simpler - highlighting the differences and maintaining the joint functionality.

Very happy to have a pass at this so we can compare and make a decision - worth exploring before going for a third schema 👍

src/export/digitalPlanning/index.ts
jessicamcinchak: Appreciate you coming back to this one ! There's an explicit reason I haven't gone with an additional argument from the planx-new side which is the (admittedly awkward) reality that discretionary/undefined-application-type payloads are also getting routed through this function to create a generic "application.json" for submission payloads (send to email & S3).

Therefore the discriminated type exposed to planx-new wouldn't be so tidy, but have to be something like this which felt pretty meaningless! 
```
"preApp" | ApplicationTypeUnion | string | undefined
```

src/export/digitalPlanning/model.ts
jessicamcinchak: Super nice idea :bookmark:

PR #632, opened 2025-01-22
fix: CVE-2024-21538

DafyddLlyr:
Fixes https://github.com/theopensystemslab/planx-core/security/dependabot/11

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,21 +51,21 @@
     "@formatjs/intl-listformat": "^7.7.9",
     "@mui/base": "5.0.0-beta.60",
     "@mui/material": "^5.15.10",
-    "ajv": "^8.17.1",
     "ajv-formats": "^2.1.1",
+    "ajv": "^8.17.1",
     "cheerio": "1.0.0",
     "copyfiles": "^2.4.1",
     "docx": "^9.1.1",
     "eslint": "^8.57.1",
     "fast-xml-parser": "^4.5.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -80,10 +80,8 @@
     "@typescript-eslint/eslint-plugin": "^7.0.0",
     "@typescript-eslint/parser": "^6.21.0",
     "@vitest/coverage-istanbul": "2.1.8",
-    "esbuild-jest": "^0.5.0",
-    "esbuild": "^0.24.2",
```
DafyddLlyr, line 84: These aren't used (as far as I can tell!).

The build works without them - it's likely they should have been removed when we swapped out jest for vite.

package.json
```diff
@@ -95,7 +93,9 @@
   "packageManager": "pnpm@8.6.6",
   "pnpm": {
     "overrides": {
-      "braces@<3.0.3": ">=3.0.3"
+      "braces@<3.0.3": ">=3.0.3",
+      "cross-spawn@>=7.0.0 <7.0.5": ">=7.0.5",
+      "semver@>=6.0.0 <6.3.1": ">=6.3.1"
```
DafyddLlyr, line 98: `pnpm audit --fix` also picked up the issue with `semver` that's not yet flagged by dependabot.

jamdelion: ooh didn't know about this command

PR #637, opened 2025-01-28
feat: Calculate exemptions

DafyddLlyr:
## What does this PR do?
 - Centralises the logic for exemptions into the `getFeeBreakdown()` function
 - The initial implementation of this had this split across this function and display layer in planx-new
 - Implemented in https://github.com/theopensystemslab/planx-new/pull/4067

Notion documentation to sit alongside this feature can be found here - https://www.notion.so/opensystemslab/How-displaying-the-fee-breakdown-works-19035d469ad180e3bcd2eab977a39a56


src/types/feeBreakdown.ts
```diff
@@ -2,13 +2,22 @@ export interface FeeBreakdown {
   amount: {
     calculated: number;
     payable: number;
-    reduction: number;
     vat: number;
-  };
+  } & ReductionOrExemption;
```
jessicamcinchak, line 6: This FeeBreakdown type is really helpful, but I do find a bit confusing that all properties appear required - whereas my instinct is that everything besides `payable` should have a `?` ?

src/utils/feeBreakdown.ts
```diff
@@ -51,6 +78,16 @@ const calculateVAT = (data: PassportFeeFields) => {
...
 
+const getReductionOrExemptionLists = (data: PassportFeeFields) => {
+  let reductions = getGranularKeys(data, "application.fee.reduction");
+  const exemptions = getGranularKeys(data, "application.fee.exemption");
+
+  // These are mutually exclusive - any exemptions invalidate reductions
+  if (exemptions.length) reductions = [];
+
+  return { reductions, exemptions };
+};
```
jessicamcinchak, line 89: Utility functions and logic are feeling very clear throughout here :+1:

jessicamcinchak (approved):

src/types/feeBreakdown.ts
DafyddLlyr: The reason for this is that I'm explicitly defaulting values to 0 when parsing the passport. 

It felt more explicit to me to have properties such as `vat: 0` as opposed to `vat: undefined`. The intention would be to say "we've checked this and it's zero" vs "this isn't included".

Functionally, as 0 is falsy we can still write `if (!feeBreakdown.vat)` etc.

jessicamcinchak: Thanks for clarification here! 

Let's keep thinking on the language for this - for example my instinct is that "we've checked this and _it doesn't apply_ (is undefined)" is desirable (and how we'd think about this for schema validation), not "we've checked this and it's zero" (how you think about for fee breakdown math) - it's tricky !!

PR #639, opened 2025-01-31
feat: automatically publish newly created flows

jessicamcinchak:
This brings E2E helpers in sync in functionality introduced here: https://github.com/theopensystemslab/planx-new/pull/4230

### DIFF
--- a/src/requests/flow.ts
+++ b/src/requests/flow.ts
@@ -148,6 +148,7 @@ export async function createFlow(
     name: string;
     data?: object;
     status?: FlowStatus;
+    userId?: number;
   },
 ): Promise<string> {
   const response: { insert_flows_one: { id: string } } = await client.request(
@@ -189,6 +190,7 @@ export async function createFlow(
       id: response.insert_flows_one.id,
       data: args.data,
     },
+    publisherId: args.userId,
     summary: "Created flow",
   });
   return response.insert_flows_one.id;
--- a/src/requests/flow.ts
+++ b/src/requests/flow.ts
@@ -184,12 +184,23 @@ export async function createFlow(
   await createAssociatedOperation(client, {
     flowId: response.insert_flows_one.id,
   });
+  await publishFlow(client, {
+    flow: {
+      id: response.insert_flows_one.id,
+      data: args.data,
+    },
+    summary: "Created flow",
+  });
   return response.insert_flows_one.id;
 }
 
 export async function publishFlow(
   client: GraphQLClient,
-  args: { flow: { id: string; data: object }; publisherId: number },
+  args: {
+    flow: { id: string; data?: object };
+    publisherId?: number;
+    summary?: string;
+  },
 ): Promise<number> {
   const response: { insert_published_flows_one: { id: number } } =
     await client.request(
@@ -207,6 +218,7 @@ export async function publishFlow(
           flow_id: args.flow.id,
           data: args.flow.data,
           publisher_id: args.publisherId,
+          summary: args.summary,
         },
       },
     );
--- a/src/requests/flow.ts
+++ b/src/requests/flow.ts
@@ -17,6 +17,7 @@ export class FlowClient {
     name: string;
     data?: object;
     status?: FlowStatus;
+    userId: number;
   }): Promise<string> {
     return createFlow(this.client, args);
   }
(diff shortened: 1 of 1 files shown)


src/requests/flow.ts
```diff
@@ -148,48 +149,63 @@ export async function createFlow(
     name: string;
     data?: object;
     status?: FlowStatus;
+    userId: number;
```
DafyddLlyr, line 152: We can also infer this from the JWT making the request - this might both avoid a lot of the associated cascading changes from adding this new argument, as well as ruling out an incorrect value being passed in.

If we just use `"x-hasura-user-id"` as a column preset, this will get read automatically from the JWT.

image

I'm pretty sure we can also toggle off access to the `publisher_id` column once this is configured as well to make sure no incorrect values get written in.

Docs: https://hasura.io/docs/2.0/auth/authorization/permissions/column-presets/

src/requests/flow.ts
```diff
@@ -148,48 +149,63 @@ export async function createFlow(
...
-          }
+  const response: { insert_flows_one: { id: string; data: object } } =
+    await client.request(
+      gql`
+        mutation CreateFlow(
+          $teamId: Int!
+          $flowSlug: String!
+          $flowName: String!
+          $data: jsonb
+          $status: flow_status_enum_enum
```
RODO94, line 163: Is it right that `enum` is stated twice in this type? Is it unfortunate naming or a typo?

RODO94: Actually sorry, just checked `user_roles` and see it's unfortunate naming

RODO94 (approved):
Looks good to me, we're adding the Publish as part of creating a flow, with a summary note stating that. I'll look through the `planx-new` change as well.

I only had one comment on the variable naming.

DafyddLlyr (commented):
Something to take a look at here which might simplify a lot of the changes in https://github.com/theopensystemslab/planx-new/pull/4230

One thing I'm not too sure of without a deep dive, is if there are places in e2e tests where the `$admin` client is creating flows as opposed to a user with a JWT (therefore no `x-hasura-user-id` session variable). If this is the case, we can insert using the `$api` role instead which relied on a JWT 👍

src/requests/flow.ts
jessicamcinchak: This is a good shout - here's my understanding of how this is currently handled: 
- `e2e/tests/api-driven`
  - `globalHelpers` all using `$admin` client (create user, create team, create flow) which is why we're manually passing around this `userId`
  - `demo-workspace` seems to be only exception to this which has uniquely defined/duplicated it's own `createFlow` mutaton. Strangely/incorrectly though, this is also the only variation of `createFlow` I've found which doesn't create an associated operation and therefore I haven't added initial publishing support either (the tests never cover either of these scenarios)
- `e2e/tests/ui-driven` 
  - `/helpers/context` similarly all using `$admin` client (create user, create team, create flow, update settings/integrations, etc)
  - There is a `generateAuthenticationToken` method in this subdir to make a JWT, but the only reference I can find to this is when we call it once via `await browserContext.addCookies()` - the future actions like creating a flow, updating flow status etc are still ultimately routed through an `$admin` client afaik

Column preset definitely feels like the right direction of travel here, but also feels like an expansive piece of tech debt bigger than publishing alone ??

src/requests/flow.ts
```diff
@@ -148,48 +149,63 @@ export async function createFlow(
...
-        insert_flows_one(
-          object: {
-            team_id: $teamId
-            slug: $flowSlug
-            name: $flowName
-            data: $data
-            version: 1
-            status: $status
-          }
+  const response: { insert_flows_one: { id: string; data: object } } =
```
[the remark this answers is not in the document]
jessicamcinchak: I simply extended the existing `response` pattern used throughout this file because I was only adding a single variable, either syntax would ultimately accomplish the same thing :+1:

src/requests/flow.ts
DafyddLlyr: Yeah similarly just been looking into this!

The column preset is a great solution for all real-life applications, but doesn't marry up to how we're creating flows in e2e environments (set up using `$admin` with no permissions, no `user` record associated via JWT).

I don't think making the `userId` param optional (for actual users) and conditionally reading argument or parsing from the JWT would make things any simpler - quite the opposite.

Agreed - this is a nice direction but a bigger change than I anticipated - this is a good route for now for sure ✅

DafyddLlyr (approved):

PR #643, opened 2025-02-05
chore: rename "placeholder" tag to "customisation"

jessicamcinchak:
Per yesterday's templates call decision !


DafyddLlyr (approved):
Just a heads up that there's a reference to this here as well - https://github.com/theopensystemslab/planx-new/blob/main/editor.planx.uk/src/pages/FlowEditor/components/Flow/components/Tag.tsx#L15

PR #655, opened 2025-02-21
fix: update Uniform descriptions to match ODP Schema

jessicamcinchak:
See thread https://opensystemslab.slack.com/archives/C089LL9LR5H/p1740137672736809

We previously updated ODP Schema mappings via https://github.com/theopensystemslab/planx-core/pull/552?email_source=slack, but failed to bring Uniform in alignment - which BOPS has pointed out !

### DIFF
--- a/src/export/oneApp/model.ts
+++ b/src/export/oneApp/model.ts
@@ -158,29 +158,21 @@ export class OneAppPayload {
           "common:IsUseChange": this.stringToBool(
             this.passport.string(["uniform.isUseChange"]),
           ),
-          "common:ProposedUseDescription": this.passport.string([
-            "proposal.description",
-          ]),
-          "common:ExistingUseDescription": this.passport.string([
-            "proposal.description",
-          ]),
+          "common:ProposedUseDescription": this.getProposalDescription(),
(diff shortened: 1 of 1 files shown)


src/export/oneApp/model.ts
```diff
@@ -380,6 +368,23 @@ export class OneAppPayload {
     };
   }
 
+  private getProposalDescription(): string {
+    // LDC-E & LDC-P descriptions are built up based on the following fields
+    const existingOrProposedDescription = [
+      this.passport.string(["proposal.buildingOperations.details"]),
+      this.passport.string(["proposal.changeOfUse.details"]),
+      this.passport.string(["proposal.existingWorks.details"]),
+      this.passport.string(["proposalexistingUse.details"]),
```
DafyddLlyr, line 377: Typo?
```suggestion
      this.passport.string(["proposal.existingUse.details"]),
```

DafyddLlyr (approved):

src/export/oneApp/model.ts
jessicamcinchak: Good spot thanks - committed !

PR #661, opened 2025-02-24
chore: refactor Planning Constraints types for single source of truth

jessicamcinchak:
Adds an exported `const` `activePlanningConstraints` which replaces duplicated, very similar instances in planx-new - see https://github.com/theopensystemslab/planx-new/pull/4372

This ensures that in the future when new Planning Data layers are added, it's a single place to update !

### DIFF
--- a/src/types/planning-constraints.ts
+++ b/src/types/planning-constraints.ts
@@ -1,3 +1,10 @@
+export type PlanningConstraintCategory =
+  | "Ecology"
+  | "Flooding"
+  | "General policy"
+  | "Heritage and conservation"
+  | "Trees";
+
 export type MinimumDigitalLandEntity = {
   "entry-date": string;
   "start-date": string;
@@ -28,7 +35,7 @@ export type Constraint = {
   value: boolean;
   text?: string;
   data?: Array<Record<string, unknown>>; // `Array<MinimumDigitalLandEntity | MinimumOSRoadFeature>` in future??
-  category?: string;
+  category?: PlanningConstraintCategory;
 };
 
 export type Metadata = {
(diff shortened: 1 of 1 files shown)


src/types/planning-constraints.ts
```diff
@@ -27,8 +34,8 @@ export type Constraint = {
   fn: string;
   value: boolean;
   text?: string;
-  data?: Array<Record<string, unknown>>; // `Array<MinimumDigitalLandEntity | MinimumOSRoadFeature>` in future??
-  category?: string;
+  data?: Array<Record<string, unknown>>; // @todo `Array<MinimumDigitalLandEntity | MinimumOSRoadFeature>`
+  category?: string; // `PlanningConstraintCategory`
```
jessicamcinchak, line 38: Setting this as the `PlanningConstraintCategory` string enum created a lot of tedious MUI List/Accordion type issues in planx-new because this prop serves as the header of each sublist. So it's a plain `string` for now, one to come back to sometime!

jamdelion (approved):

src/types/planning-constraints.ts
```diff
@@ -58,3 +65,198 @@ export interface GISResponse {
 export interface EnhancedGISResponse extends GISResponse {
   planxRequest: string;
 }
+
+export interface BasePlanningConstraintSchema {
+  name: string;
+  neg: string;
```
RODO94, line 71: question: is the `neg` and `pos` types here a good use case for Template Literals? 

Something like: 
```
neg: `${'is not'}${string}`
```

To ensure when people add these constraints you have a positive and negative option, or is this too restrictive and overkill?

RODO94 (approved):
I went through this and cross-referenced it with the planx-new changes. I couldn't find anything that was mis-typed here that would cause issues there. Especially looking at the `Object.entries()` in planx-new and how they correspond to the shape of the data here.

I really like moving this to planx-core as well, makes it more readable and maintainable for me.

src/types/planning-constraints.ts
jessicamcinchak: The phrasing of these is quite custom so I don't think worth trying to template them! Eg `is not` versus `is not, or is not within,...` and singular versus plural constraint names.

PR #675, opened 2025-03-05
feat: add materials mappings for schema

jessicamcinchak:
Per request from Meera at Lambeth - this one is a great example of a type we already had/have in the schema, but have never brought our passport mappings up to sync with ! We need to do this more pro-actively at some point !

See https://opendigitalplanning.slack.com/archives/C08BL63KCKB/p1741175213567829

### DIFF
--- a/src/export/digitalPlanning/mocks/planningPermission.ts
+++ b/src/export/digitalPlanning/mocks/planningPermission.ts
@@ -15,6 +15,18 @@ export const mockPlanningPermissionSession = {
   sanitised_at: null,
   email: "example@example.com",
   passport: {
+    "proposal.materials": [
+      {
+        type: "window",
+        existing: "Glaze window",
+        proposed: "Glaze window",
+      },
+      {
+        type: "wall",
+        existing: "Red bricks",
+        proposed: "Brown bricks",
+      },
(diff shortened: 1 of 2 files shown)


src/export/digitalPlanning/mocks/planningPermission.ts
```diff
@@ -15,6 +15,18 @@ export const mockPlanningPermissionSession = {
...
+        type: "window",
+        existing: "Glaze window",
+        proposed: "Glaze window",
+      },
+      {
+        type: "wall",
+        existing: "Red bricks",
+        proposed: "Brown bricks",
+      },
+    ],
```
jessicamcinchak, line 29: This is a snippet direct from Meera's example. At some point we'll want to fully regenerate all these mocks, but piecemeal additions are working okay in the meantime. 

My process for this is:
- Add mock `passport` data values
- Temporarily edit `src/export/digitalPlanning/model.test.ts` to log out properties of interest
Screenshot from 2025-03-05 13-28-25
- Run `pnpm test` & confirm logs are as expected 
Screenshot from 2025-03-05 13-26-20

DafyddLlyr (approved):

PR #689, opened 2025-03-25
chore: Rename is_copiable column

DafyddLlyr:


src/requests/flow.ts
```diff
@@ -416,31 +419,34 @@ async function setStatus(
...
-        isCopiable: args.isCopiable,
+        canCreateFromCopy: args.canCreateFromCopy,
       },
     );
 
     return flow;
   } catch (error) {
     new Error(
-      `Failed to update flow visibility to "${args.isCopiable}". Error: ${error}`,
+      `Failed to update flow visibility to "${args.canCreateFromCopy}". Error: ${error}`,
```
jessicamcinchak, line 449: nit: can we pass `args.flow.id` into this error message too ?

jessicamcinchak (approved):

src/requests/flow.ts
DafyddLlyr: Good catch thanks - added!

PR #694, opened 2025-04-02
feat: add `serviceCharge`, `fastTrack` and total `vat` amounts to fee breakdown

jessicamcinchak:
Supports https://github.com/theopensystemslab/planx-new/pull/4508

Use https://4508.planx.pizza/testing/set-fee-demo for passport mocks

### DIFF
--- a/src/types/feeBreakdown.ts
+++ b/src/types/feeBreakdown.ts
@@ -32,5 +32,5 @@ export interface PassportFeeFields {
   "application.fee.serviceCharge.VAT"?: number;
   "application.fee.fastTrack"?: number;
   "application.fee.fastTrack.VAT"?: number;
-  // `${string}.VAT`?: string;
+  [key: `${string}.VAT`]: number | undefined;
 }
--- a/src/types/feeBreakdown.ts
+++ b/src/types/feeBreakdown.ts
@@ -2,7 +2,9 @@ export interface FeeBreakdown {
   amount: {
     calculated: number;
     payable: number;
-    vat: number;
+    vat?: number;
+    serviceCharge?: number;
+    fastTrack?: number;
   } & ReductionOrExemption;
   reductions: string[];
   exemptions: string[];
@@ -21,10 +23,14 @@ export type ReductionOrExemption =
 export interface PassportFeeFields {
   "application.fee.calculated": number;
   "application.fee.payable": number;
-  "application.fee.payable.includesVAT": boolean;
   "application.fee.reduction.alternative": boolean;
   "application.fee.reduction.parishCouncil": boolean;
   "application.fee.reduction.sports": boolean;
   "application.fee.exemption.disability": boolean;
   "application.fee.exemption.resubmission": boolean;
+  "application.fee.serviceCharge"?: number;
+  "application.fee.serviceCharge.VAT"?: number;
+  "application.fee.fastTrack"?: number;
+  "application.fee.fastTrack.VAT"?: number;
+  // `${string}.VAT`?: string;
 }
--- a/src/utils/feeBreakdown.test.ts
+++ b/src/utils/feeBreakdown.test.ts
@@ -74,6 +74,25 @@ describe("calculateReductionOrExemption() helper function", () => {
     expect(reduction).toEqual(0);
     expect(exemption).toEqual(100);
   });
+
+  it("correctly outputs a 0 reduction when payable is greater than calculated due to a service charge", () => {
+    const input: PassportFeeFields = {
+      "application.fee.calculated": 100,
+      "application.fee.payable": 136,
+      "application.fee.serviceCharge": 30,
+      "application.fee.serviceCharge.VAT": 6,
+      "application.fee.reduction.alternative": false,
+      "application.fee.reduction.parishCouncil": false,
+      "application.fee.reduction.sports": false,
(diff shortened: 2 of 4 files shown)


src/types/feeBreakdown.ts
```diff
@@ -21,10 +23,14 @@ export type ReductionOrExemption =
...
   "application.fee.reduction.alternative": boolean;
   "application.fee.reduction.parishCouncil": boolean;
   "application.fee.reduction.sports": boolean;
   "application.fee.exemption.disability": boolean;
   "application.fee.exemption.resubmission": boolean;
+  "application.fee.serviceCharge"?: number;
+  "application.fee.serviceCharge.VAT"?: number;
+  "application.fee.fastTrack"?: number;
+  "application.fee.fastTrack.VAT"?: number;
+  // `${string}.VAT`?: string;
```
jessicamcinchak, line 35: Haven't gotten this dynamic type working yet - see this skipped test with example key ! https://github.com/theopensystemslab/planx-core/pull/694/files#diff-935434782beb09636d1698cb7deddb0d16142667ffb111db3736ca2284aeb77cR114

src/utils/feeBreakdown.test.ts
```diff
@@ -129,7 +182,27 @@ describe("toFeeBreakdown() helper function", () => {
...
+      "application.fee.reduction.sports": false,
+      "application.fee.exemption.disability": true,
+      "application.fee.exemption.resubmission": false,
+    };
+
+    const { amount } = toFeeBreakdown(input);
+
+    expect(amount.payable).toEqual(input["application.fee.payable"]);
+    expect(amount.vat).toEqual(input["application.fee.serviceCharge.VAT"]);
+    expect(amount.exemption).toEqual(input["application.fee.calculated"]);
```
jessicamcinchak, line 205: In order for the exemption to correctly show here, I had to swap the order of key lookup here https://github.com/theopensystemslab/planx-core/pull/694/files#diff-441beea3935905070627d583371c0f78ef44ee5f5fc7fb2e9d89ecfc707f86f4R55 (no other tests impacted by this change - so think it was a legitmate oversight/inconsistency!)

src/utils/feeBreakdown.ts
```diff
@@ -122,7 +125,11 @@ export const createPassportSchema = () => {
     .object({
       "application.fee.calculated": feeSchema.optional().default(0),
       "application.fee.payable": feeSchema,
-      "application.fee.payable.includesVAT": booleanSchema,
+      "application.fee.serviceCharge": feeSchema.optional().default(0),
+      "application.fee.serviceCharge.VAT": feeSchema.optional().default(0),
+      "application.fee.fastTrack": feeSchema.optional().default(0),
+      "application.fee.fastTrack.VAT": feeSchema.optional().default(0),
+      // `${z.string()}.VAT`: feeSchema.optional().default(0),
```
jessicamcinchak, line 132: Similar to above - expect to be able to specify a dynamic `.VAT` property here but haven't landed on correct syntax yet !

src/types/feeBreakdown.ts
DafyddLlyr, line 35:
```suggestion
  [key: `${string}.VAT`]: number | undefined;
```

I think it's just the `[key: x]` syntax missing here 👍

DafyddLlyr: Yep - just for clarity it's needed there also. The tests just cover the helper function `sumVAT()` - if we were to add other `.vat` variables into the actual passport (any input into `getFeeBreakdown()` it would fail as these would be stripped away as additional keys).

PR https://github.com/theopensystemslab/planx-core/pull/695 will resolve this 👍

src/types/feeBreakdown.ts
```diff
@@ -21,10 +23,14 @@ export type ReductionOrExemption =
...
   "application.fee.reduction.alternative": boolean;
   "application.fee.reduction.parishCouncil": boolean;
   "application.fee.reduction.sports": boolean;
   "application.fee.exemption.disability": boolean;
   "application.fee.exemption.resubmission": boolean;
+  "application.fee.serviceCharge"?: number;
+  "application.fee.serviceCharge.VAT"?: number;
+  "application.fee.fastTrack"?: number;
+  "application.fee.fastTrack.VAT"?: number;
+  [key: `${string}.VAT`]: number | undefined;
```
DafyddLlyr, line 35:
```suggestion
  "application.fee.serviceCharge": number;
  "application.fee.serviceCharge.VAT": number;
  "application.fee.fastTrack": number;
  "application.fee.fastTrack.VAT": number;
  [key: `${string}.VAT`]: number;
```

I think this is actually correct as the Zod schema is converting these to 0 - `feeSchema.optional().default(0)`

jessicamcinchak: This was my original instinct also - but when required, existing tests "inputs" will error that there are missing fields - eg 
Screenshot from 2025-04-02 18-02-22

Therefore, I think optional will ensure that the fee breakdown doesn't break for existing invite to pay sessions (passport fields already written to `payment_requests`, but actual page/fee breakdown not going to be opened/displayed for up to 28 days) when this is merged?

jessicamcinchak:
Closing in favor of #696 - lots of dynamic type comments here no longer apply, we've simplified our approach !

PR #696, opened 2025-04-04
feat: add `fastTrack`, `serviceCharge`, `paymentProcessing` fees and granular VAT to fee breakdown

jessicamcinchak:
Fee breakdown updates required to support implementation of a new `SetFee` component type in the editor https://github.com/theopensystemslab/planx-new/pull/4550

**Changes:**
- Removes `application.fee.payable.includesVAT` and `amount.vat`as these were never used in "real" content and do match actual implementation
- Removes `calculateVAT()` helper function as this is now handled in the editor directly and the fee breakdown will simply receive pre-calculated passport variables which reflect applicable VAT
- Outputs `amount.exemption` & `amount.reduction` as _difference_ (so often negative) to more clearly communicate edge cases of postive reductions (aka increases or modifications)
- Updates many tests throughout to have more realistic mock scenarios that include new VAT-able extra charges


src/types/feeBreakdown.ts
```diff
@@ -20,8 +27,15 @@ export type ReductionOrExemption =
...
+  "application.fee.calculated.VAT"?: number;
   "application.fee.payable": number;
-  "application.fee.payable.includesVAT": boolean;
+  "application.fee.payable.VAT"?: number;
+  "application.fee.fastTrack"?: number;
+  "application.fee.fastTrack.VAT"?: number;
+  "application.fee.serviceCharge"?: number;
+  "application.fee.serviceCharge.VAT"?: number;
+  "application.fee.paymentProcessing"?: number;
+  "application.fee.paymentProcessing.VAT"?: number;
```
DafyddLlyr, line 38: This doesn't match the type generated by our `staticSchema`.

As we're importing a unknown passport and then outputting `feeSchema.optional().default(0)` the resulting types will always be `number` not `number | undefined`.

DafyddLlyr: We could also infer this type from the `staticSchema` in order to keep them in sync - 

```ts
type PassportFeeFields = z.infer<typeof staticSchema>
```

src/types/feeBreakdown.ts
```diff
@@ -1,8 +1,15 @@
...
+    calculatedVAT?: number;
     payable: number;
-    vat: number;
+    payableVAT?: number;
+    fastTrack?: number;
+    fastTrackVAT?: number;
+    serviceCharge?: number;
+    serviceChargeVAT?: number;
+    paymentProcessing?: number;
+    paymentProcessingVAT?: number;
```
DafyddLlyr, line 12: Same as previous comments, these would be just `number` not `number | undefined`.

src/utils/feeBreakdown.ts
```diff
@@ -98,44 +105,55 @@ const getReductionOrExemptionLists = (data: PassportFeeFields) => {
...
+const feeSchema = z.union([questionSchema, setValueSchema]).transform(toNumber);
+
+/** Describes how boolean values are set via PlanX components */
+const booleanSchema = z
+  .tuple([z.enum(["true", "false"])])
+  .default(["false"])
+  .transform(toBoolean);
+
+/** Static fee-associated passport fields */
+export const staticSchema = z.object({
```
DafyddLlyr, line 133: Now that we don't have the separation of static vs dynamic, I'm not sure if this name still helps - is just `schema` or `passportFeeFieldsSchema` better?

src/utils/feeBreakdown.ts
```diff
@@ -98,44 +105,55 @@ const getReductionOrExemptionLists = (data: PassportFeeFields) => {
...
+  "application.fee.calculated": feeSchema.optional().default(0),
+  "application.fee.calculated.VAT": feeSchema.optional().default(0),
+  "application.fee.payable": feeSchema, // only required fee and inclusive of VAT
+  "application.fee.payable.VAT": feeSchema.optional().default(0), // sum of all VAT
+  "application.fee.fastTrack": feeSchema.optional().default(0),
+  "application.fee.fastTrack.VAT": feeSchema.optional().default(0),
+  "application.fee.serviceCharge": feeSchema.optional().default(0),
+  "application.fee.serviceCharge.VAT": feeSchema.optional().default(0),
+  "application.fee.paymentProcessing": feeSchema.optional().default(0),
+  "application.fee.paymentProcessing.VAT": feeSchema.optional().default(0),
```
DafyddLlyr, line 143: Given the number of `.optional().default()` here vs the single required fee, it might be clearer to capture these as variables?

```ts
const optinalFeeSchema = z.union([questionSchema, setValueSchema]).transform(toNumber).optional().default(0);
const requiredFeeSchema = z.union([questionSchema, setValueSchema]).transform(toNumber);
```

src/utils/feeBreakdown.ts
```diff
@@ -98,44 +105,55 @@ const getReductionOrExemptionLists = (data: PassportFeeFields) => {
...
+  "application.fee.exemption.resubmission": booleanSchema,
+});
 
+/**
+ * Generate schemas, parse passport, then transform to a FeeBreakdown object
+ */
 export const getFeeBreakdown = (passportData: unknown): FeeBreakdown => {
-  const schema = createPassportSchema();
-  const result = schema.parse(passportData);
+  const staticResult = staticSchema.parse(passportData);
```
DafyddLlyr, line 155: nit: I think there's a merge conflict here with the `safeParse()` PR - https://github.com/theopensystemslab/planx-core/pull/705

src/utils/feeBreakdown.ts
```diff
@@ -47,41 +45,50 @@ const getCalculatedAmount = (data: PassportFeeFields) =>
 export const calculateReductionOrExemptionAmounts = (
   data: PassportFeeFields,
 ): ReductionOrExemption => {
-  const hasExemption =
-    data["application.fee.exemption.disability"] ||
-    data["application.fee.exemption.resubmission"];
+  const { reductions, exemptions } = getReductionOrExemptionLists(data);
+
+  const hasExemption = exemptions.length > 0;
```
DafyddLlyr, line 50: Nice and future proof!

src/utils/feeBreakdown.ts
```diff
@@ -47,41 +45,50 @@ const getCalculatedAmount = (data: PassportFeeFields) =>
...
   if (hasExemption) {
     return {
-      exemption:
-        data["application.fee.payable"] || data["application.fee.calculated"],
+      exemption: -getCalculatedAmount(data),
```
DafyddLlyr, line 53: 👍

src/utils/feeBreakdown.ts
```diff
@@ -47,41 +45,50 @@ const getCalculatedAmount = (data: PassportFeeFields) =>
...
+      reduction: 0,
+    };
+  }
+
+  // Reductions should exclude and be calculated prior to extra VAT-able charges & fees being applied
+  const extraCharges =
+    (data["application.fee.serviceCharge"] || 0) +
+    (data["application.fee.fastTrack"] || 0) +
+    (data["application.fee.paymentProcessing"] || 0) +
+    (data["application.fee.payable.VAT"] || 0); // sum of all VAT
```
DafyddLlyr, line 71: The ` || 0`s should be redundant here as we're setting this as the default value.

src/utils/feeBreakdown.ts
```diff
@@ -47,41 +45,50 @@ const getCalculatedAmount = (data: PassportFeeFields) =>
...
 
-  // A negative reduction indicates a content issues with passport variables
-  // "application.fee.calculated" should always be greater than "application.fee.payable"
-  if (reduction < 0) throw Error("Reduction should always be negative");
+  // A negative reduction indicates a possible content issue with passport variables
+  // "application.fee.calculated" should be greater than "application.fee.payable"
+  //   except in possible edge cases of sports club flat fee reduction/modification which may be higher than certain application fees
+  const hasSportsReduction = reductions.includes("sports");
+  if (!hasSportsReduction && reduction > 0)
+    throw Error("Non-sports reductions expected to be negative");
```
DafyddLlyr, line 84: More specific and meaningful - great!

DafyddLlyr (changes requested):
Looking great - clear and easy to follow as always 👍 Appreciate the additional test coverage and better mocks.

A few small type issues and merge conflicts to take a look at, no showstoppers!

src/types/feeBreakdown.ts
jessicamcinchak: Can I just confirm that it's actually safe to make these _required_ and still expect _existing_ sessions to work when they reach a Pay component and try to render a breakdown?

When required, many now-"partial" test inputs fail on "missing properties", but these mocks feel like accurate indicators of what we should expect _existing_ session data to look like - eg 
```
const input: PassportFeeFields = {
      "application.fee.calculated": 100,
      "application.fee.payable": 50,
      "application.fee.payable.includesVAT": false,
      "application.fee.reduction.alternative": false,
      "application.fee.reduction.parishCouncil": false,
      "application.fee.reduction.sports": false,
      "application.fee.exemption.disability": false,
      "application.fee.exemption.resubmission": false,
    };
```

DafyddLlyr: Yes, this will account for existing sessions.

Actual passports come into `getFreeBreakdown()` as an `unknown` type, which will parse them into the known and safe `PassportFeeFields` type. So if a passport does not have a value for `"application.fee.serviceCharge"` (very likely!) it will still always end up as a `PassportFeeFields` with `"application.fee.serviceCharge": 0`.

The issue you're describing here is the unit tests for the private helper functions which will expect a full `PassportFeeFields` type (already generated by `staticSchema.parse(passportData)`). 

We can solve this in a few ways I think
 - Run all tests through `getFreeBreakdown()` 
   - Probably "best" from a testing POV not to test private methods, but it's super awkward and not a pattern we generally stick to 
 - Use `Partial<T>` and casting to make TS happy
 - Use full `PassportFeeFields` mocks which is what will actually be passed to the helper functions being tested

DafyddLlyr (approved):

src/utils/feeBreakdown.ts
jessicamcinchak: Now rebased :heavy_check_mark:

PR #699, opened 2025-04-07
feat: Parse responses from `AddressInput` schema fields

DafyddLlyr:
This PR adds validation for the `AddressInput` field added to the schema / list component in PlanX.

### DIFF
--- a/src/export/bops/utils/index.ts
+++ b/src/export/bops/utils/index.ts
@@ -7,6 +7,7 @@ import {
   QuestionAndResponses,
   Response,
 } from "../../../types/index.js";
+import { SchemaResponses } from "./schema.js";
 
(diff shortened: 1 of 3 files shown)


src/export/bops/utils/index.ts
```diff
@@ -61,12 +62,6 @@ export const formatQuestion = ({
     ? fieldTitle
     : `[${schemaType} ${index + 1}] ${fieldTitle}`;
 
-export type SchemaResponses =
```
DafyddLlyr, line 64: Previously, this was repeated - we can actually infer this from the existing Zod schemas which should mean fewer changes in future.

jessicamcinchak (approved):

PR #705, opened 2025-04-10
chore: Use Zod's `safeParse()` instead of `parse()` for better error handling and formatting

DafyddLlyr:
## What does this PR do?
 - Uses Zod's `safeParse()` instead of `parse()` to allow us to `.flatten()` (format) errors into a more readable format
   - The `safeParse()` pattern seems a little cleaner than wrapping `try/catch` blocks everywhere
   - The default Zod error (a list of issues) is comprehensive, but hard to read - `.flatten()` gives a nicer format we can more easily match to the input object
   - Docs: https://zod.dev/ERROR_HANDLING?id=formatting-errors

## Context
This steps from the following discussion - https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1744009225907769 (OSL Slack)

### DIFF
--- a/src/utils/feeBreakdown.test.ts
+++ b/src/utils/feeBreakdown.test.ts
@@ -338,7 +338,19 @@ describe("getFeeBreakdown() function", () => {
         "some.other.fields": ["abc", "xyz"],
       };
 
-      expect(() => getFeeBreakdown(mockPassportData)).toThrow();
+      expect(() => getFeeBreakdown(mockPassportData)).toThrowError(
+        expect.objectContaining({
+          message: expect.stringMatching(/Failed to parse fee breakdown data/),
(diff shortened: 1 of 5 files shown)


src/utils/feeBreakdown.test.ts
DafyddLlyr, line 1: @jessicamcinchak I've updated these tests to demonstrate the structure of the error thrown. If this contradicts or breaks anything in https://github.com/theopensystemslab/planx-core/pull/696 we can keep them as just plain `toThrows()`.

jessicamcinchak: Thanks for flagging - these should be fine / I'll handle rebasing 696 (it's also open for review)

jessicamcinchak (approved):

PR #712, opened 2025-04-16
feat: Toggle flows offline when setting team status to trial

DafyddLlyr:
Implemented in https://github.com/theopensystemslab/planx-new/pull/4588

Further details can be found here - https://github.com/theopensystemslab/planx-new/pull/4588/files#r2046563181

### DIFF
--- a/src/requests/team.ts
+++ b/src/requests/team.ts
@@ -445,22 +445,31 @@ async function updateTeamSettings(
     gql`
       mutation UpdateTeamSettings(
         $team_id: Int
-        $teamSettings: team_settings_set_input!
+        $team_settings: team_settings_set_input!
+        $toggle_flows_offline: Boolean!
       ) {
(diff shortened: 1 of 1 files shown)


src/requests/team.ts
```diff
@@ -445,22 +445,31 @@ async function updateTeamSettings(
...
             team_id
             id
           }
         }
+
+        # If a team is marked as "trial", toggle all flows offline
+        update_flows(
+          where: { team_id: { _eq: $team_id } }
+          _set: { status: offline }
+        ) @include(if: $toggle_flows_offline) {
```
DafyddLlyr, line 465: This `@include` directive is new to me and seems pretty handy!

Docs: https://hasura.io/docs/2.0/queries/postgres/variables-aliases-fragments-directives/#using-directives

PR #737, opened 2025-05-19
fix: CVE-2025-46565

DafyddLlyr:
Fixes https://github.com/theopensystemslab/planx-core/security/dependabot/24

### DIFF
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -5,6 +5,7 @@ export default defineConfig({
     testTimeout: 10_000,
     globals: true,
     environment: "node",
+    pool: "threads",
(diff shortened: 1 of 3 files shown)


vitest.config.ts
```diff
@@ -5,6 +5,7 @@ export default defineConfig({
     testTimeout: 10_000,
     globals: true,
     environment: "node",
+    pool: "threads",
```
DafyddLlyr, line 8: Fix found here - https://github.com/vitest-dev/vitest/discussions/6511

jessicamcinchak (approved):

PR #748, opened 2025-06-04
fix: correct redaction for single-line `ContactInput` transformation in `proposal_details`

jessicamcinchak:
See #help-issues report from Gateshead here https://opendigitalplanning.slack.com/archives/C0241GWFG4B/p1749033657201129


DafyddLlyr (approved):
Thanks for the fix here - appreciate the extra test coverage!

PR #795, opened 2025-08-25
fix: Handle incorrectly normalised payments

DafyddLlyr:
Implemented in https://github.com/theopensystemslab/planx-new/pull/5119, please see PR for full context.

### DIFF
--- a/src/templates/html/application/ApplicationHTML.tsx
+++ b/src/templates/html/application/ApplicationHTML.tsx
@@ -28,8 +28,9 @@ function Highlights(props: { data: PlanXExportData[] }): JSX.Element {
   )?.responses as GovUKPayment | undefined;
   const payRef = govPayPayment?.payment_id;
 
-  const fee = props.data.find((d) => d.question === "application.fee.payable")
+  const feeInPence = props.data.find((d) => d.question === "payment_amount")
     ?.responses as number | undefined;
+  const fee = feeInPence ? feeInPence / 100 : undefined;
 
   return (
     <Box component="dl" sx={{ ...gridStyles, border: "none" }}>


src/templates/html/application/ApplicationHTML.tsx
```diff
@@ -27,7 +27,15 @@ function Highlights(props: { data: PlanXExportData[] }): JSX.Element {
...
   const payRef = govPayPayment?.payment_id;
-  const fee = govPayPayment?.amount;
+
+  // Temp check to account for incorrectly normalised payment amounts
+  // refund_summary is reliable as this is derived from an audit snapshot taken immediately as the payment was made and will reflect the full amount paid
+  const feeInPence = govPayPayment?.refund_summary?.amount_available;
+  const fee = feeInPence ? feeInPence / 100 : undefined;
+
+  // Revert to the below line after after 01/10/25
+  // const fee = govPayPayment?.amount
```
DafyddLlyr, line 37: Approximate time to review and merge + 28 day session expiry period.

I've set a Slack reminder to revisit this and tidy up come October.

jessicamcinchak: Based on above, I think we could just straight up replace this and not do the 28 day circle-back dance?

jessicamcinchak (approved):

src/templates/html/application/ApplicationHTML.tsx
```diff
@@ -27,7 +27,15 @@ function Highlights(props: { data: PlanXExportData[] }): JSX.Element {
     (d) => d.question === "application.fee.reference.govPay",
   )?.responses as GovUKPayment | undefined;
   const payRef = govPayPayment?.payment_id;
-  const fee = govPayPayment?.amount;
+
+  // Temp check to account for incorrectly normalised payment amounts
+  // refund_summary is reliable as this is derived from an audit snapshot taken immediately as the payment was made and will reflect the full amount paid
+  const feeInPence = govPayPayment?.refund_summary?.amount_available;
+  const fee = feeInPence ? feeInPence / 100 : undefined;
```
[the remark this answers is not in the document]
DafyddLlyr: Much simpler thanks, now updated!

We don't have `application.fee.payable` directly here, but we have `payment_amount` which is mapped here into the CSV data - 

https://github.com/theopensystemslab/planx-core/blob/42f00a0e9bf1d405af278808ebc7e18a9013548f/src/export/bops/index.ts#L487-L492

I've updated this and then planx-new and tested, and I'm getting the expected output in `overview.html` 

Image

PR #802, opened 2025-08-28
chore: Drop document templates from OneApp payload generation

DafyddLlyr:


jessicamcinchak:
I think all of these changes are a duplicate of https://github.com/theopensystemslab/planx-core/pull/801 ?

PR #812, opened 2025-09-08
feat: generate an HTML document with a map and legend table based on `MapAndLabel` output

jessicamcinchak:
Implemented in https://github.com/theopensystemslab/planx-new/pull/5189

### DIFF
--- a/src/templates/html/mapAndLabel/MapAndLabelHTML.tsx
+++ b/src/templates/html/mapAndLabel/MapAndLabelHTML.tsx
@@ -10,7 +10,6 @@ export function MapAndLabelHTML(props: {
   drawColor: string;
   schemaFieldValues: string[];
   schemaName: string;
-  osApiKey?: string;
 }) {
   return (
     <html>
@@ -35,7 +34,6 @@ export function MapAndLabelHTML(props: {
           features={props.geojson}
           clipGeojsonData={props.boundingBox}
           drawColor={props.drawColor}
-          osApiKey={props.osApiKey}
         />
         <Table
           geojson={props.geojson}
--- a/src/templates/html/mapAndLabel/MapAndLabelMap.tsx
+++ b/src/templates/html/mapAndLabel/MapAndLabelMap.tsx
@@ -70,7 +70,6 @@ export default function MapAndLabelMap(props: {
   features: object;
   clipGeojsonData: GeoJSON.Feature;
   drawColor: string;
-  osApiKey?: string;
 }) {
   return (
     <>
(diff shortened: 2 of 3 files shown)


src/templates/html/mapAndLabel/MapAndLabelHTML.tsx
```diff
@@ -0,0 +1,80 @@
...
+          border-collapse: collapse;
+          margin-top: 20px;
+          min-width: 650px;
+          max-width: 800px;
+        }
+        th,
+        td {
+          padding: 10px;
+          word-wrap: break-word;
+        }
```
jessicamcinchak, line 76: Table styling uses very basic CSS - suggestions welcome !

src/templates/html/mapAndLabel/MapAndLabelMap.tsx
```diff
@@ -0,0 +1,96 @@
...
+      <my-map
+        id="map-and-label-map"
+        showNorthArrow={true}
+        hideResetControl={true}
+        showScale={true}
+        useScalebarStyle={true}
+        showPrint={true}
+        collapseAttributions={true}
+        basemap={props.osApiKey ? `OSVectorTile` : `OSM`}
+        osApiKey={props.osApiKey || ""}
```
jessicamcinchak, line 86: In local testing, I couldn't manage to get `osProxyEndpoint` working (I guess because of CORS?) so I was passing the OS API key directly. 

Any better ideas here? When generating an HTML _document_, is there any alternative to bundling the key directly? Should we store it as an env var in planx-core and read in that way, or pass env var from planx-new?

src/templates/mocks/mapAndLabelMocks.ts
```diff
@@ -0,0 +1,212 @@
...
+        species: "Oak",
+        work: "Trim branches",
+        reason:
+          "Some branches are overgrowing into the neighbours garden beyond my property curtilege",
+      },
+    },
+  ],
+};
+
+export const mapAndLabelNodePropsA: any = {
```
jessicamcinchak, line 58: I've intentionally mocked "NodeProps" to match the full new breadcrumb data I'm submitting in planx-new, even though we ultimately only need to rely on a couple of these props to generate the doc ! Makes it a little easier to iterate / understand universe of what's available.

DafyddLlyr:
Now that https://github.com/theopensystemslab/planx-new/pull/5208 is merged to staging, we can switch out the `osAPIKey` for the `osProxyEndpoint` prop here 👍

src/templates/html/mapAndLabel/MapAndLabelMap.tsx
jessicamcinchak: Now solved via https://github.com/theopensystemslab/planx-new/pull/5208 - thanks @DafyddLlyr !

Going to pick up switch to `osProxy` in a follow-up PR :heavy_plus_sign:

jessicamcinchak: Now updated to use `osProxyEndpoint` instead ! Still pointing at `api.editor.planx.dev` because https://github.com/theopensystemslab/planx-new/pull/5208 isn't on prod yet - we can come back and adjust this later !

PR #832, opened 2025-10-01
fix: build application from JSON (instead of CSV)

jmcbroom:
start to convert `ApplicationHTML.tsx` to use the JSON data structure instead of the .csv

this goes along with the `jimmy/build-html-from-json` [branch in planx-new](https://github.com/theopensystemslab/planx-new/compare/jimmy/build-html-from-json?expand=1)

### DIFF
--- a/src/templates/html/application/ApplicationHTML.tsx
+++ b/src/templates/html/application/ApplicationHTML.tsx
@@ -90,8 +90,7 @@ function Result(props: { data: Application }): JSX.Element {
 }
 
 function AboutTheProperty(props: { data: Application }): JSX.Element {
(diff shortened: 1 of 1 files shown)


src/templates/html/application/ApplicationHTML.tsx
```diff
@@ -147,7 +142,7 @@ function Boundary(props: { data: PlanXExportData[] }): JSX.Element {
 }
 
 function ProposalDetails(props: {
-  data: PlanXExportData[];
+  data: QuestionAndResponses[];
```
jessicamcinchak: This line is a low-key very meaningful type change / `PlanXExportData[]` is a real legacy quirk :raised_hands: :broom:

PR #967, opened 2026-04-10
chore(ci): Set min release age to match planx-new

DafyddLlyr:
Setup to match https://github.com/theopensystemslab/planx-new/pull/6400


jessicamcinchak (approved):
Thanks for this one!

PR #982, opened 2026-04-29
fix: condition 'Paid' wording on pay reference

augustlindemer:
We want to avoid suggesting we know whether or not a fee was already paid if we don't have a GOV pay reference, e.g. due to external payment. This conditions the 'paid' wording in the overview.htm on payRef, defaulting to 'payable' or 'submitted' wording if we don't.


DafyddLlyr (approved):

jessicamcinchak (approved):
Smart, thank you ! And no underlying ODP Schema property names change here, so shouldn't break anything downstream for folks using Power Automate etc ➕

PR #992, opened 2026-05-09
fix: ODP Schema `metadata.service.enhancements` expects object not array

jessicamcinchak:
Debugging a recently failed submission! 

- Our payload mapping generated an array of objects: https://api.editor.planx.uk/admin/session/f923cd54-7441-44ac-8e21-f37bf4739b50/digital-planning-application?skipValidation=true
Screenshot from 2026-05-09 17-08-37

- While the schema expects an optional object: https://theopensystemslab.github.io/digital-planning-data-schemas-docs/
Screenshot from 2026-05-09 17-10-34

### DIFF
--- a/src/export/digitalPlanning/mocks/planningPermission.ts
+++ b/src/export/digitalPlanning/mocks/planningPermission.ts
@@ -15,13 +15,16 @@ export const mockPlanningPermissionSession = {
   sanitised_at: null,
   email: "example@example.com",
   passport: {
-    "_enhancements": {
+    _enhancements: {
       "proposal.description": {
-        "enhanced": "Roof extension to the rear of the property, incorporating starship launchpad.",
-        "original": "Really big roof addition off the back of the house for starship launchpad."
-      }
+        enhanced:
+          "Roof extension to the rear of the property, incorporating starship launchpad.",
+        original:
+          "Really big roof addition off the back of the house for starship launchpad.",
+      },
     },
-    "enhancedTextInput.proposal.description.action": "Accepted the AI-enhanced description",
+    "enhancedTextInput.proposal.description.action":
+      "Accepted the AI-enhanced description",
     "proposal.parking": [
       {
         type: "cars",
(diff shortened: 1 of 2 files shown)


src/export/digitalPlanning/mocks/planningPermission.ts
```diff
@@ -15,6 +15,16 @@ export const mockPlanningPermissionSession = {
...
+    _enhancements: {
+      "proposal.description": {
+        enhanced:
+          "Roof extension to the rear of the property, incorporating starship launchpad.",
+        original:
+          "Really big roof addition off the back of the house for starship launchpad.",
+      },
+    },
+    "enhancedTextInput.proposal.description.action":
+      "Accepted the AI-enhanced description",
```
jessicamcinchak, line 27: Have also updated mock session data here ! 

These full mocks are tedious and slow to regenerate in full, but truthfully even just mocking _passport only_ (eg no associated breadcrumb, linked mock flow does _not_ actually have EnhancedTextInput component type etc) is still enough in short term to better catch validation errors here !

src/export/digitalPlanning/model.ts
```diff
@@ -1488,24 +1488,22 @@ export class DigitalPlanning {
...
+            enhanced:
+              this.passport.data?.["_enhancements"]?.["proposal.description"]?.[
+                "enhanced"
+              ],
+            userAction:
+              this.passport.data?.[
+                "enhancedTextInput.proposal.description.action"
+              ],
+            model: "Google Gemini",
+          },
```
jessicamcinchak, line 1506: `enhancements` plural probably _should_ be an `[{}]` in future (especially when we have more than just project description). But that will require a schema change + release which is much more timely to coordinate. 

In order to resubmit quickly, we want to adjust our payload generation mappings only here. AFAIK BOPS isn't reading this metadata yet anyways !

DafyddLlyr (approved):

PR #1008, opened 2026-06-08
fix: ensure PlanningDesignation has JSON Schema `$id`

jessicamcinchak:
Need to fix this properly in next ODP Schema release, but can patch in meantime here to enable `property.constraints.planning` auto-suggest in editor

### DIFF
--- a/src/export/digitalPlanning/schemas/application/schema.json
+++ b/src/export/digitalPlanning/schemas/application/schema.json
@@ -7728,6 +7728,7 @@
       "title": "Planning constraint"
     },
     "PlanningDesignation": {
+      "$id": "PlanningDesignation",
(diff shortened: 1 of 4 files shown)


src/export/digitalPlanning/schemas/application/schema.json
```diff
@@ -7728,6 +7728,7 @@
       "title": "Planning constraint"
     },
     "PlanningDesignation": {
+      "$id": "PlanningDesignation",
```
jessicamcinchak, line 7731: This is the fundamental change! 

Other file changes here are auto-output from running `pnpm generate-types-from-json` and linter

jessicamcinchak:
Not the solution to my bug!

PR #1009, opened 2026-06-08
chore: ensure `getValidSchemaValues` can handle `PlanningDesignation` union

jessicamcinchak:
Necessary to auto-suggest data values for `property.constraints.planning` in editor!


freemvmt (approved):
looks good!

PR #1014, opened 2026-06-18
chore: refactor `sortBreadcrumbs` to sort on `seq` instead of depth-first

jessicamcinchak:
Follows on from https://github.com/theopensystemslab/planx-new/pull/6844

Todo: 
- [x] Update `sortBreadcrumbs`
- [x] Manually update small test mocks to add `seq`
- [ ] Regenerate all full-service flow + session data mocks (eg `/src/export/digitalPlanning/mocks`) to reflect `seq`
  - Honestly is this worth it?? We could refactor function to sort on `seq` if exists (will apply to all "real" data) or keep depth-first traversal if `seq` undefined (will apply to all "mock" data) ?? Obviously makes tests a lot less useful/thorough here, but not an easy or quick task to regenerate across flows/services !!

### DIFF
--- a/src/models/session/logic.ts
+++ b/src/models/session/logic.ts
@@ -70,36 +70,9 @@ export function sortBreadcrumbs(
   flow: FlowGraph,
   breadcrumbs: Breadcrumbs,
 ): OrderedBreadcrumbs {
-  const breadcrumbMap = new Map(Object.entries(breadcrumbs));
-  const breadcrumbsInOrder = new Map<NodeId, Crumb>();
-
-  const visited = new Set<NodeId>();
-  const stack: string[] = [...(flow._root.edges || [])].reverse();
-
-  while (stack.length > 0) {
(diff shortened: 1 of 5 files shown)


src/models/session/logic.ts
```diff
@@ -70,36 +70,9 @@ export function sortBreadcrumbs(
...
-    // Node edges are traversed left-to-right
-    // We process our stack in last-in, first-out order
-    // This means we need to iterate backwards over edges
-    for (let i = currentNode.edges.length - 1; i >= 0; i--) {
-      const childId = currentNode.edges[i];
-      if (!visited.has(childId)) {
-        stack.push(childId);
-      }
-    }
-  }
```
jessicamcinchak, line 102: We can remove this linear stack traversal which is great! 

But we'll still always require a `flow` argument in this version of `sortBreadcrumbs` because we find/assign each crumb's `sectionId` in the returned shape.

PR #1015, opened 2026-06-22
chore: Update dependabot issues

DafyddLlyr:
This should resolve the following 5 issues - https://github.com/theopensystemslab/planx-core/security/dependabot

Once approved and merged, I'll update planx-new which will resolve a bunch of issues there.


pnpm-workspace.yaml
```diff
@@ -1 +1,3 @@
-minimumReleaseAge: 10080
\ No newline at end of file
+minimumReleaseAge: 10080
+minimumReleaseAgeExclude:
+ - undici@8.5.0
```
DafyddLlyr: Temp override, pinned to an exact version which will be 7 days old later today. Depending on when this PR gets approved I'll remove this exclusion.

I'd love to find a proper working solution to this - seems pretty tedious to have to manually manage these.

package.json
```diff
@@ -96,6 +96,11 @@
   "pnpm": {
     "onlyBuiltDependencies": [
       "esbuild"
-    ]
+    ],
+    "overrides": {
+      "undici@>=7.23.0 <7.28.0": ">=7.28.0",
```
freemvmt: think this line is just a subset of the next one - superfluous?

freemvmt (approved):
🔧

pnpm-workspace.yaml
DafyddLlyr: Removed today as we're now passed the 7 day timeout

package.json
```diff
@@ -55,7 +55,7 @@
     "@emotion/react": "^11.14.0",
     "@emotion/styled": "^11.14.1",
     "@formatjs/intl-listformat": "^8.3.8",
-    "@mui/material": "^9.0.0",
+    "@mui/material": "<9.1.0",
```
DafyddLlyr, line 58: Range changed to avoid this issue - https://github.com/mui/material-ui/issues/48636

PR #1018, opened 2026-06-30
chore: allow for errors and optional enhancements in AI-generated descriptions

zz-hh-aa:
So that in cases where AI was not able to generate an enhanced description, the user will not fail validation.

### DIFF
--- a/src/export/digitalPlanning/model.test.ts
+++ b/src/export/digitalPlanning/model.test.ts
@@ -441,7 +441,7 @@ describe("DigitalPlanning", () => {
       );
     });
 
-    it("should include error when enhancement fails", () => {
+    it("should include error in 'enhanced' property when enhancement fails", () => {
       const mock = mockSessions[5];
       const instance = new DigitalPlanning({
         sessionId: "c06eebb7-6201-4bc0-9fe7-ec5d7a1c0797",
@@ -464,12 +464,13 @@ describe("DigitalPlanning", () => {
       const payload = instance.getPayload();
 
       const applicationMetadata = payload.metadata as PlanXMetadata;
+
       expect(applicationMetadata.service!.enhancements!).toHaveProperty(
-        "error",
+        "enhanced",
         "INVALID_INPUT",
       );
       expect(applicationMetadata.service!.enhancements!).not.toHaveProperty(
-        "enhanced",
+        "error",
       );
     });
   });
--- a/src/export/digitalPlanning/model.test.ts
+++ b/src/export/digitalPlanning/model.test.ts
@@ -29,6 +29,7 @@ import {
 import { mockPriorApprovalSession } from "./mocks/priorApproval.js";
 import { mockReportAPlanningBreachSessionMedway } from "./mocks/reportAPlanningBreach.js";
 import { DigitalPlanning } from "./model.js";
+import { PlanXMetadata } from "./schemas/application/types.js";
 
 // `getPlanningConstraints` relies on an accurate teamSlug to be available, other vars can be be mocked
 const mockMetadataForSession = (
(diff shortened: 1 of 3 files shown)


src/export/digitalPlanning/schemas/application/schema.json
```diff
@@ -2945,13 +2948,8 @@
...
       },
-      "required": [
-        "dataProperty",
-        "original",
-        "enhanced",
-        "userAction",
-        "model"
-      ],
+      "required": ["dataProperty", "original", "userAction", "model"],
+      "anyOf": [{ "required": ["enhanced"] }, { "required": ["error"] }],
```
zz-hh-aa: Should this be one or the other, rather than `anyOf`? To handle this specific scenario we only need either / or, but I wasn't sure if we wanted this to be flexible enough to be able to validate a case where an enhanced description was generated and there was an error?

src/export/digitalPlanning/model.test.ts
```diff
@@ -416,4 +417,60 @@ describe("DigitalPlanning", () => {
       });
     });
   });
+
+  describe("enhancements", () => {
```
zz-hh-aa, line 421: I kept this outside of the other test block because virtually the same application (minus `enhanced` property) didn't need to be iterated through in all of the other tests.

src/export/digitalPlanning/schemas/application/schema.json
jessicamcinchak: So something important to understand when editing schemas: 
- The `schema.json` files found in planx-core come from https://github.com/theopensystemslab/digital-planning-data-schemas/tree/dist (manually copied latest version in lieue of the schema being published/import-able from NPM or similar). 
- Editing here in planx-core is _not_ the same as editing the schema and making a new release! Modifications made here would alter our internal validation only but _not_ be available to other schema users which would cause drift and issues ! 
- Instead, let's find a "mapping"-only solution that still works within the existing schema version's published types - does that make sense? 
  - So for example, if we need to work around `"enhanced"` being required metadata, can we simply write enhanced _or_ error message to this field?

src/export/digitalPlanning/model.test.ts
jessicamcinchak: Great!

jessicamcinchak (changes requested):
Thanks again for offering to pick this one up and get into the weeds of very-long `model.ts` with us, left a comment below - I think solution here is actually even simpler than what you've started! Happy to discuss tmrw too if useful!

src/export/digitalPlanning/schemas/application/schema.json
zz-hh-aa: Ah thanks for the context!! I think I've understood but if I haven't let's catch up about it later. 

In [cd03d47](https://github.com/theopensystemslab/planx-core/pull/1018/commits/cd03d476dda6ac695fc4c6a793c303fa1c22bb83) I've updated `model.ts` to write `error` to `enhanced` now, so an error should show in the payload as:
```
{
    dataProperty: 'proposal.description',
    original: 'Test',
    enhanced: 'INVALID_INPUT',
    userAction: 'Accepted the AI-enhanced description',
    model: 'Google Gemini'
  }
```

(Previous schema changes also undone in [f75a306](https://github.com/theopensystemslab/planx-core/pull/1018/commits/f75a30666cd54768a772ec117fa7ce823560d222))

jessicamcinchak (approved):
Changes look great, thanks for added test coverage !

src/export/digitalPlanning/schemas/application/schema.json
jessicamcinchak: Yes exactly - thank you ! 

Since this is just the `metadata` section, we can be quite unfussed about `enhanced` just showing the error! 

What's important, and what councils/back-office systems will actually read, is `data.proposal.description` which should be displaying "original" :slightly_smiling_face:

PR #1024, opened 2026-07-03
fix: Move `react` and `react-dom` from `dependencies` to `peerDependencies`

DafyddLlyr:
## What's the problem?
`planx-core` renders React  e.g. `generateApplicationHTML()`), but has `react` and `react-dom` declared as regular `dependencies`. That makes `planx-core` bundle its own copy of React. When a consuming app resolves React to even a slightly different patch version, two copies end up in the dependency tree (verified with `pnpm why react` in `planx-new`).

Downstream in the API this this surfaced as a crash during HTML generation - 

```
Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
See https://react.dev/link/invalid-hook-call for tips about how to debug and fix this problem.
```

### What's the solution?
From the [docs](https://react.dev/warnings/invalid-hook-call-warning#duplicate-react) - 

> _If you see more than one React, you’ll need to figure out why this happens and fix your dependency tree. For example, maybe a library you’re using incorrectly specifies react as a dependency (rather than a peer dependency)._

- Move `react` and `react-dom` to `peerDependencies` so `planx-new` will dedupe onto it's React copy
- Add `react` and `react-dom` to `devDependencies` so `planx-core` still builds and can be tested locally


freemvmt (approved):
Nice fix

jessicamcinchak (approved):
Helpful explanation, thanks for fixing ! Do we expect MUI to hit same issue?

DafyddLlyr:
> _Do we expect MUI to hit same issue?_

Great question - hadn't considered that. I expect the same fix would be wise here - let me read up and I'll open another PR to confirm this.

PR #1027, opened 2026-07-10
fix: don't throw error on missing session preview keys

jessicamcinchak:
See thread https://opensystemslab.slack.com/archives/C4B0CKQ3U/p1783504433807739

Airbrake has been logging a number of 
```
Error: could not initiate a payment request: passport key "proposal.projectType" not found in passport data
```
despite removing this requirment from publish validation checks and simply allowing ITP emails and summary page to fallback to "Not submitted". We need to similarly loosen this requirement here.

There are now too many discretionary fee-carrying services to reasonably enforce these summary keys.

### DIFF
--- a/src/requests/payment-request.test.ts
+++ b/src/requests/payment-request.test.ts
@@ -23,31 +23,6 @@ describe("extractSessionPreviewData", () => {
     );
   });
 
-  test("keys must be present in the passport", () => {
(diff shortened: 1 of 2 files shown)


DafyddLlyr (approved):
Thanks for the fix + explainer!

PR #1039, opened 2026-07-24
fix: ensure `payment_requests.session_preview_data` has correct types

jessicamcinchak:
See this bug thread https://opensystemslab.slack.com/archives/C088K9ZL8EA/p1784553514231659

Regression/bug introduced in https://github.com/theopensystemslab/planx-core/pull/1027


DafyddLlyr (approved):
Thanks for additional test coverage here also 👌

PR #1045, opened 2026-08-04
Revert "fix: handle "No result" flags in ODP Schema `preAssessment` mappings"

jessicamcinchak:
Reverts theopensystemslab/planx-core#1044

This temporarily validated a broken submission, but we want to continue throwing these errors to fix broken content journeys!


DafyddLlyr (approved):

jessicamcinchak:
Merging this to `main` but intentionally not immediately bumping `planx-new`, can be picked up next time ! We want this particular session to continue validating against original "fix" during Camden's download window.

